PR #1, opened 2020-12-21
Styles-based comments, license files, tests

kezhenxu94:


LICENSE
```diff
@@ -199,3 +199,21 @@
...
+licenses.
+
+========================================================================
+MIT licenses
+========================================================================
+
+The following components are provided under the MIT license. See project link for details.
+The text of each license is also included at licenses/LICENSE-[project].txt.
+
+    languages.yaml from GitHub: https://github.com/github/linguist/blob/master/lib/linguist/languages.yml MIT
```
kezhenxu94, line 219: @wu-sheng please recheck this, I copy this file from GitHub's linguist repo

wu-sheng: From the license's perspective, I think it is fine. About the paper work, I think you miss `licenses/LICENSE-linguist.txt`?

wu-sheng: Is this related to the runtime? Could you explain a little about how we use this?

kezhenxu94: This file contains a full list of mappings between file types and their extensions, we parse this at runtime to get the mapping and map the extensions to our comment styles.

This could be downloaded at runtime, but that would be troublesome for people who have poor/no network access

kezhenxu94: Containing this file in our binary makes it more self-contained IMO

wu-sheng: > This file contains a full list of mappings between file types and their extensions, we parse this at runtime to get the mapping and map the extensions to our comment styles.

Then this should be included in source codes license and dist license.

> This could be downloaded at runtime, but that would be troublesome for people who have poor/no network access

No need, I was just checking from license perspective.

kezhenxu94:
> Then this should be included in source codes license and dist license.

I think I did it (in riot LICENSE and dist/LICENSE)

PR #8, opened 2020-12-24
Add Contact Us for the 1st release.

wu-sheng:


README.adoc
```diff
@@ -17,8 +17,18 @@
...
 = SkyWalking Eyes
 
+<img src="http://skywalking.apache.org/assets/logo.svg" alt="Sky Walking logo" height="90px" align="right" />
+
 Infra tools that are created and used by the SkyWalking Team.
 
+[![Twitter Follow](https://img.shields.io/twitter/follow/asfskywalking.svg?style=for-the-badge&label=Follow&logo=twitter)](https://twitter.com/AsfSkyWalking)
+
+# Contact Us
+* Submit [an issue](https://github.com/apache/skywalking/issues/new) by using [INFRA] as title prefix.
```
kezhenxu94, line 27: Actually I plan to enable the issues tracker in this repo individually , this repo is slightly different from the other sub projects, it’s not much related to skywalking itself, it’s better to have the issues here rather than in the main repo

wu-sheng: This would be a little tricky, take a look at our website. There is a requirement from ASF which need to post an issue report channel on there. How should we separate this?

kezhenxu94: image

Do you mean this? Can we have multiple links

wu-sheng: Yes, this one. Technically, we can. But do we really need that? From my understanding, unless this is going to become another TLP, a separate issue tracker seems not very necessary. After all, still our PMC members, like you and @fgksgf , are taking care of the repo.

kezhenxu94 (approved):

PR #10, opened 2020-12-24
Add script to build release tars and move licenses to belonging subproject

kezhenxu94:


kezhenxu94:
Just moved the licenses into the belonging subproject and will release the subproject individually

wu-sheng:
Just a reminder, as a monorepo, you will face the case, all modules have to be ready for a release at the same time. Otherwise, you will break the consistency of version for different models. This lead me to think another issue, do we really need monorepo. The only reason would be we will release multiple binary(s) at once.

PR #11, opened 2020-12-26
Shape the quality-eye sub-directory for NGE2E

kezhenxu94:
This patch creates a sub-directory to host the new E2E framework codes, @fgksgf @Humbertzhang , let's host the codes here instead of in the CLI repo, they are independent after all.

@wu-sheng this also demonstrates how we organize different tools in this repo, we may have "release-eye" (checking gpg / sha512, etc.) in the future that stands alongside "license-eye", "quality-eye", etc.. Hope the structures make sense to you

### DIFF
--- a/quality-eye/README.md
+++ b/quality-eye/README.md
@@ -1,0 +1,5 @@
+# Quality Eye - NGE2E
+
+NGE2E is the next generation End-to-End Testing framework that aims to help developers to set up, debug, and verify E2E tests with ease. It’s built based on the lessons learnt from tens of hundreds of test cases in the SkyWalking main repo.
+
+**[Design doc](https://skywalking.apache.org/blog/e2e-design/).**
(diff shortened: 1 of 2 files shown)


wu-sheng:
I don't have any concern about the structure from the beginning. The thing I wanted to point out is it will become a factor to determine when to release.
Because your git tag needs to match the apache source tar. But as the monorepo, it is very hard, or impossible.

wu-sheng:
We definitely could have multiple tools in every release, but basically, please make sure all tools are available in every release(git tag).

quality-eye/README.md
```diff
@@ -0,0 +1,5 @@
+# Quality Eye - NGE2E
+
+NGE2E is the next generation End-to-End Testing framework that aims to help developers to set up, debug, and verify E2E tests with ease. It’s built based on the lessons learnt from tens of hundreds of test cases in the SkyWalking main repo.
```
wu-sheng, line 3: From the validation part, https://skywalking.apache.org/blog/e2e-design/#verify, this is not a common usage e2e test tool.
Do you plan to expand the scope?

kezhenxu94: [This](https://github.com/apache/skywalking-website/pull/184) should address your question

PR #12, opened 2020-12-26
 Init the structure of quality-eye

fgksgf:
- change `README.adoc` to `README.md` in license-eye
- init the structure of quality-eye by following https://github.com/golang-standards/project-layout/


wu-sheng:
@kezhenxu94 I think we need to required check and reviewer approve option to make the PR update automatically.

wu-sheng:
Then, @kezhenxu94 @fgksgf @Humbedooh,what is the plan for 0.1.0 release? I remember it should include license-eye only.

wu-sheng:
This PR is listed in the 0.1.0 milestone, and codes of `quality-eye` are going to be added. I want to be clear how the release plan. Because we need time to market this repo and tool.

fgksgf:
Maybe we need two milestones, one for `license-eye` and the other for `quality-eye` ?

wu-sheng:
Take a look about this https://github.com/apache/skywalking-eyes/pull/10#issuecomment-750902397.

As a monorepo, you have to make sure all codes are ready, no matter how many binary tars you are going to provide.
You could say 0.1.0 including the license-eye tool, and 0.2.0 for both license-eye and quality-eye. 

The key thing is, git tag must be ready and as one whole package as one version release.

wu-sheng:
> Maybe we need two milestones, one for license-eye and the other for quality-eye ?

0.1.0-licence-eye and 0.1.0-quality-eye would be very confused as both of them includes uncompleted codes if I check out the tag.

wu-sheng:
@kezhenxu94 From this PR status, the required CI task is not working.

kezhenxu94:
> @kezhenxu94 From this PR status, the required CI task is not working.

Only after when you receive an email (something like below) sent by ASF GitBox can it take effect. But we didn't receive until now

image

wu-sheng:
But, I can't see anything works.

kezhenxu94:
> But, I can't see anything works.

I've asked the infra team

kezhenxu94:
> > But, I can't see anything works.
> 
> I haven't receive the email yet. I've asked the infra team.

Worked

kezhenxu94:
> As a monorepo, you have to make sure all codes are ready

1. I don't think so, we can release only parts of the codes in this repo. The parts we actually released will go into https://dist.apache.org

> Maybe we need two milestones, one for `license-eye` and the other for `quality-eye` ?

2. I consider this is a good idea (`license-eye-0.1.0` and `quality-eye-0.1.0`), this allows us to track the corresponding issues/PRs in each tool.

> 0.1.0-licence-eye and 0.1.0-quality-eye would be very confused as both of them includes uncompleted codes if I check out the tag.

3. Because of (1), we will use another release method, that is, check out a new branch for releasing (only pick out what will be really released, for example, `license-eye` directory), and tag upon that branch, when voting, we use the tag and the commit id of that branch, there is no different steps as what we've been doing already. And there is no "uncompleted codes if you check out the tag".

As this is a brand new repo and the 2 tools (license-eye and quality-eye) are at their early development, IMO (3) allows us to move fast and won't block each other in terms of codes consistency or readiness.

@wu-sheng any further concern?

wu-sheng:
If you set two branches, why need one repo?  For me, I am not following the ideas why making the project so complex.
From my understanding, making two things ready before every release should be fine, as far as this moment. If we reached that days, I prefer to separate the repo.

wu-sheng:
I haven't seen any project using branch to release partial project. Does any of you have?
New repo is not limited in the ASF or SkyWalking. If the way is not typical and widely in oss, we should not consider this.

kezhenxu94:
> If you set two branches, why need one repo? For me, I am not following the ideas why making the project so complex.

I personally don't see any complexity of using a release branch. All things work as before except for the releasing preparation work, which only involves with the release manager.

> From my understanding, making two things ready before every release should be fine, as far as this moment. If we reached that days, I prefer to separate the repo.

"making two things ready before every release" is fine but not easy, especially when you have more tools in the future, I'm not sure whether I'll create any new tool though. Especially when the tool only contains a small amount of codes, I don't REALLY want to create and maintain a repo for that.

> I haven't seen any project using branch to release partial project. Does any of you have? If the way is not typical and widely in oss, we should not consider this.

There are many, the one you are most familiar with is DolphinScheduler. Others like Istio, Kubernetes, etc.. 

> New repo is not limited in the ASF or SkyWalking. 

I know that, @fgksgf and @Humbertzhang are willing to maintain another new repo for the E2E, I'm ok to create one.

This PR will be closed and the codes should be pushed to the newly-created repo, I'll just reshape this repo to only host license-eye tool.

kezhenxu94:
@fgksgf please delete the branch after you backed up the changes locally

wu-sheng:
Istio isn't. They maintain release branches, yes. But they are same main stream.
You are talking about two totally separate things.

wu-sheng:
The complexity is not about tech, it is logical. Who will use a main branch for a monorepo, but two releases are including the 50%?

PR #13, opened 2020-12-26
Fix strict is not working

wu-sheng:


.asf.yaml
```diff
@@ -28,3 +28,8 @@ github:
     main:
       required_status_checks:
         strict: true
+        contexts:
+          - LicenseEye
```
wu-sheng, line 32: @kezhenxu94 If no required task, the `strict: true` doesn't work.

.asf.yaml
```diff
@@ -28,3 +28,8 @@ github:
     main:
       required_status_checks:
         strict: true
+        contexts:
+          - build-license-eye
```
kezhenxu94, line 32: This is not right

.asf.yaml
```diff
@@ -28,3 +28,8 @@ github:
...
+      required_pull_request_reviews:
+        dismiss_stale_reviews: true
+        required_approving_review_count: 1
```
kezhenxu94, line 35: I don't want to enable this in this repo.

https://github.com/apache/skywalking-python/pull/107#discussion_r548336881

We enabled it in the main repo because we faced problem that committer(s) merged the PRs too early, but in these repos where we have limited committers and I think we just leave it as is until we really need, so that some trivial changes can be merged as quickly as possible.

wu-sheng: Personally, I am not sure what is an trivial change. Mostly, you have more than one committer,they need to sync. 
Reviewing doesn't have to be finding bug or code quality.

.asf.yaml
wu-sheng: What should put here? I thought it  should be task name.

.asf.yaml
kezhenxu94: Enabling "update branch" doesn't require reviews actually, it only requires checks to pass, which I agree to enable it, also with that enabled, no one can push to the master branch directly

PR #16, opened 2020-12-27
Fix wrong path of Dockerfile in action.yaml

kezhenxu94:


wu-sheng (approved):

kezhenxu94:
Verified by https://github.com/apache/skywalking-python/pull/110/checks?check_run_id=1613578753 good for release now

fgksgf (approved):

PR #17, opened 2020-12-31
Finalize 0.1, fix a minor bug, add docs, etc.

kezhenxu94:


README.md
```diff
@@ -41,9 +41,7 @@ header:
 
 **NOTE**: The full configurations can be found in [the configuration section](#configurations).
 
-### Use as a Binary
-
-#### Install
+### Compile from Source
```
kezhenxu94, line 44: We will have another 2 sections: "Use with Docker image", "Use as a Binary" after we have the release binary and docker image

CHANGES.md
```diff
@@ -4,3 +4,7 @@
   + Add `check` and `fix` command.
   + `check` results can be reported to pull request as comments.
   + `fix` suggestions can be filed on pull request as edit suggestions.
+
+- Dependencies License Report
+  + Support Golang `go.mod` dependency report.
+  + Licences Apache 2.0, BSD-3-Clause and MIT can be identified.
```
kezhenxu94, line 10: Also want to ask for @wu-sheng your opinion on, whether we want to include this "Dependencies License Report" in the first version, or just don't mention this in the change log until we have more complete supported list of languages / licenses, maybe next version?

wu-sheng: I prefer to skip the unstable features.

wu-sheng (approved):

PR #18, opened 2021-01-08
Identify some of dual licenses and skip built-in packages' licenses

kezhenxu94:
This PR adds a pattern to identify dual-license and skip resolving built-in package's licenses


wu-sheng (approved):
I remember you mentioned, license check should be 0.2.0 scope. Are we going to add more into 0.1.0?

kezhenxu94:
Just a minor polish, we won't highlight dependencies' license check in 0.1.0.

BTW, hopefully I can start the release process within this weekend.

PR #19, opened 2021-01-09
Fix the inaccurate description of README.md.

zifeihan:
Fix the inaccurate description of README.md, the `Compile from Source` in README.md is incorrect, and there is no license-eye directory in the skywalking-eyes directory.


kezhenxu94 (approved):
Thanks

PR #20, opened 2021-01-12
Add documentation about how to use via docker image

kezhenxu94:
The docker image is really large, still pushing....

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,8 @@
 # under the License.
 #
 
+# TODO: minimize the docker image size, now 524MB !!!
+
 FROM golang:1.14.3-alpine AS build
 
 WORKDIR /license-eye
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -16,6 +16,8 @@
 # under the License.
 #
 
+# TODO: minimize the docker image size, now 524MB !!!
+
```
wu-sheng, line 20: Haha, why so huge?

kezhenxu94: Didn't notice this before because I used alpine as base image, maybe because I use `apk` to download some dependencies inside the image, need sometime to investigate and minimize it

wu-sheng (approved):

Dockerfile
kezhenxu94: After compressed, it's 184MB, https://hub.docker.com/repository/docker/apache/skywalking-eyes/tags?page=1&ordering=last_updated

PR #24, opened 2021-03-22
feat: support ini extension

chengshiwen:
Support ini extension like https://github.com/apache/skywalking/blob/master/test/e2e/e2e-test/docker/php/php-shadow.ini


kezhenxu94 (approved):
Thanks

PR #27, opened 2021-03-31
Update NOTICE

kezhenxu94:


wu-sheng (approved):

kezhenxu94:
> I think this belongs 0.3.0

This is skywalking-eyes, we haven't have 0.2.0 yet

PR #31, opened 2021-05-11
remove duplicate check

zooltd:
the function `config.PathsIgnore` would be called twice each time a filepath is checked


kezhenxu94:
This is intended, the first check acts as a fast-return pattern.

For example, if the ignore paths include `.git/**` (or `.git`), the entire `.git` folder will be ignored without listing all files recursively under `.git` and perform ignoring logic, which saves a lot of time

PR #36, opened 2021-07-07
Fix minor error

MoGuGuai-hzr:
Reading an file named `""` will raise the `IsNotExist` error, so the conditional statement is unnecessary in `pkg/config/config.go`

The result should match whether the pattern is `.` or `./`.

Small improvement.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -37,12 +37,11 @@ type Config struct {
 func (config *Config) Parse(file string) (err error) {
 	var bytes []byte
 
-	if file != "" {
-		logger.Log.Infoln("Loading configuration from file:", file)
+	// attempt to read configuration from specified file
+	logger.Log.Infoln("Loading configuration from file:", file)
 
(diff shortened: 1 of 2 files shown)


wu-sheng:
> Reading an file named `""` will raise the `IsNotExist` error, so the conditional statement is unnecessary in `pkg/config/config.go`
> 
> The result should match whether the pattern is `.` or `./`.
> 
> Small improvement.

Thanks for the feedback, where are you going to use SkyWalking-eye tool?

MoGuGuai-hzr:
> 
> 
> > Reading an file named `""` will raise the `IsNotExist` error, so the conditional statement is unnecessary in `pkg/config/config.go`
> > The result should match whether the pattern is `.` or `./`.
> > Small improvement.
> 
> Thanks for the feedback, where are you going to use SkyWalking-eye tool?

Yes, i will. And i am a student participating Summer 2021 of Open Source Promotion Plan.

kezhenxu94 (approved):
Thank you @MoGuGuai-hzr very much 🙇🏻 nice catches!

PR #40, opened 2021-07-09
Add support for PHP

emschu:
This is a preview to add support for PHP header comments with minimal code changes. I'm not sure if it is a good idea to change the `CommentStyle` struct this way by adding two new properties `ensure_after` or `ensure_before`. 

My idea is composing the header license text by *ensure_after* + *header* + *ensure_before* - if the RegEx provided in `after` yields no results.
Are there other (script) languages which require a similar behaviour?

What do you think about this? Or do you want to discuss this in a dedicated issue?

Some notes:
- `declare(strict_types=[0|1]);` has to be omitted or it has to be the very first statement in the script, which is pretty common in modern PHP code.
- empty php files are considered
- PHP can be configured to use a short-tag syntax by enabling `short_open_tag` in `php.ini`.  `<?php` then can be written as `<?`, but even with `short_open_tag`s set to `On` `<?php` is still a valid start of a PHP code block.


wu-sheng:
@heyanlong This is for license check for PHP, welcome to provide feedback.

kezhenxu94:
@emschu hey, thanks for bringing this up.

Is it allowed to have multiple `<?php ... ?>` in a single `.php` file, in another word, is the following codes are valid in PHP?


```php
<?php
/**
 * header license comments
 */
?>

<?php
// some real codes
?>
```

If it's valid, maybe we can simple use the following comment style config,

```yaml
- id: PHP
  start: |
    <?php
    /*
  middle: ' *'
  end: |
    */
    ?>
```

multi-line strings in yaml is rather legit and normal.

emschu:
@ke4qqq Yes, your example is a valid php script syntax. You can read about this topic [here](https://www.php.net/manual/en/language.basic-syntax.phptags.php).

> If a file contains only PHP code, it is preferable to omit the PHP closing tag at the end of the file. This prevents accidental whitespace or new lines being added after the PHP closing tag, which may cause unwanted effects because PHP will start output buffering when there is no intention from the programmer to send any output at that point in the script.

For this reason I think it is not a good idea to generally add a php closing tag after the license header in PHP files as this could cause trouble with whitespace or output buffering in pure php files (*pure* means: non-template PHP code, e.g. class definitions etc.).

I will think about this problem next week again. Enjoy the weekend!

heyanlong (dismissed):

kezhenxu94 (approved):
Looks good to me now, thanks @emschu !! 🙇🏻

PR #42, opened 2021-07-13
Add support of apostrophe style comments for PlantUML

emschu:
As discussed in #41 and should/must be merged before it.


kezhenxu94 (approved):
🎉 thank you for the quick fix @emschu

PR #44, opened 2021-07-16
 fix some program compatibility issues in Windows

zooltd:
The embeded file system `embed.FS` only takes the forward slash (`/`) as the path separator. However, the operating system path separator in Windows is the backward slash (`\`). We need to normalize the file path separator each time we use `embed.FS`.


kezhenxu94 (approved):
Nice catch! We might also consider adding Windows to the CI test matrices.

PR #45, opened 2021-07-22
Add support for resolving npm dependencies' licenses

zooltd:


pkg/deps/npm.go
```diff
@@ -0,0 +1,170 @@
...
+func (resolver *NpmResolver) Resolve(packageFile string, report *Report) error {
+	deps, err := resolver.parseDeps(packageFile)
+	if err != nil {
+		return err
+	}
+
+	root := filepath.Dir(packageFile)
+	if skip := resolver.QuerySkipInstallPkgs(); !skip {
+		if err := resolver.InstallPkgs(root); err != nil {
+			logger.Log.Errorln("Fail to install packages")
```
kezhenxu94, line 63: We should return here

```suggestion
			return fmt.Errorf("fail to install packages:%+v", err)
```

pkg/deps/npm.go
```diff
@@ -0,0 +1,170 @@
...
+type NpmResolver struct {
+	Resolver
+}
+
+// Package represents package.json
+type Package struct {
+	Name         string            `json:"name"`
+	Version      string            `json:"version"`
+	License      string            `json:"license"`
+	Dependencies map[string]string `json:"dependencies"`
```
kezhenxu94, line 44: We can also capture the `license` field in this file so that we can simply use if it exists. This field is more precise than identifying from the LICENSE text, what do you think?

kezhenxu94 (commented):
Looks promising!! Left some comments

pkg/deps/npm.go
zooltd: Oh, I had intended to do that but I forgot.😓 Youn're right. This should be on the first priority when inferring the licenses.

PR #48, opened 2021-07-23
Add support for resolving npm dependencies' licenses

zooltd:
1. Parse the project package.json file to gather the required packages
2. Run command 'npm install' to install or update the required node packages (can skip)
3. Walk through each package's root directory to resolve licenses
    - STEP 1: Try to find and parse the package.json file to capture the license field
    - STEP 2: Try to find the license file to identify the license

### DIFF
--- a/pkg/deps/npm.go
+++ b/pkg/deps/npm.go
@@ -1,0 +1,219 @@
+//
+// Licensed to Apache Software Foundation (ASF) under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Apache Software Foundation (ASF) licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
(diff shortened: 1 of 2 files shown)


wu-sheng:
You should update the doc to show how to use this.

kezhenxu94 (commented):
This is a very good start point to resolve dependencies’ licenses in NPM projects. 

This PR only resolves the direct dependencies’ licenses, will you continue to resolve their transitive dependencies in next PR?

zooltd:
> This is a very good start point to resolve dependencies’ licenses in NPM projects.
> 
> This PR only resolves the direct dependencies’ licenses, will you continue to resolve their transitive dependencies in next PR?

Glad to take it. Also, I'm planning to resolve dev-dependencies' licenses.

wu-sheng:
Notice, dev-dependency is not binary level or source code level dependency.
They mostly are considered as a convenient tools to enhance dev stage.
I could recommend to process transitive dependency. A.k.a, dependencies of dependencies. Those are true runtime dependencies too.

wu-sheng:
Dev-tool should not include commercial or unknown licenses. But even GPL and AGPL are fine to use, that is what I mean different.

zooltd:
> Notice, dev-dependency is not binary level or source code level dependency.
> They mostly are considered as a convenient tools to enhance dev stage.
> I could recommend to process transitive dependency. A.k.a, dependencies of dependencies. Those are true runtime dependencies too.

Got it. I will work on this feature in next PR.

kezhenxu94 (approved):

PR #50, opened 2021-07-25
add feature: support resolving pom.xml for maven

MoGuGuai-hzr:
Solution: copy, using `mvn dependency:copy-dependencies`, all dependencies and transitive dependencies, and then resolve license from the jar files.

Known disadvantages: 
- depends on maven
- replication of the jar files needs to read and write the disk


pkg/deps/maven_test.go
```diff
@@ -0,0 +1,56 @@
+package deps_test
```
wu-sheng, line 1: About this test case, I can't see you check the expected jars list. Could you enhance this test case? And notice, `test` and `provided` scopes should be checked, both these are not dependency check scope.

wu-sheng (commented):
One question, this PR is only including dependency resolving, but not checking the dependencies license compatible, right?

kezhenxu94:
> One question, this PR is only including dependency resolving, but not checking the dependencies license compatible, right?

All the dependencies licenses resolvers including Go, NPM don’t check license compatibility now, they just resolve the licenses for human’s reference. Checking license compatibility is out of the scope of this PR and will be discussed after the main resolvers (Java, NodeJS, Golang) can correctly resolve the licenses first.

pkg/deps/maven_test.go
kezhenxu94: > And notice, `test` and `provided` scopes should be checked, both these are not dependency check scope.

This is misleading, they should NOT be checked. I edited your comment

kezhenxu94: If you check the method this PR does (using `mvn copy-dependencies`), it only copies runtime dependencies, the same way we did in main repo to check licenses dependencies.

wu-sheng (commented):
This and #48 should be very similar PR, but why do the changes have so many differences?
1. `Resolvers` is not changed here.
2. `.licensserc.yaml` is not changed in #48

pkg/deps/maven_test.go
wu-sheng: I noticed this use `mvn copy-dependencies`, I mean the test case should verify.

pkg/deps/maven.go
```diff
@@ -0,0 +1,162 @@
...
+)
+
+func init() {
+	var err error
+	_, err = exec.Command("mvn", "--version").Output()
+	if err != nil {
+		return
+	}
+
+	Resolvers = append(Resolvers, new(MavenPomResolver))
```
kezhenxu94, line 27: I think this is not a good condition to check whether we should add the resolver:

- `mvn version` is not enough to determine whether maven resolver can be use, [maven wrapper](https://github.com/takari/maven-wrapper) is also widely used in some projects.
- We should log some warning / error logs when users' environment don't have maven installed, instead of failing silently.

My suggestion is 

- just add the resolver to https://github.com/apache/skywalking-eyes/blob/d6e232b01042d112b103ae6c7f51c5c677e1f37e/pkg/deps/resolve.go#L29-L32
- check whether `mvnw` exists in the same directory as `pom.xml` first, if yes, use `mvnw`, otherwise, use `mvn`
- run the needed command (`mvn copy-dependencies`), if it failed, we also fail the command and print some logs

.licenserc.yaml
```diff
@@ -83,3 +83,4 @@ header: # `header` section is configurations for source codes license header.
 dependency:
   files:
     - go.mod
+    - test/testdata/pom.xml
```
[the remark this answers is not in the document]
MoGuGuai-hzr: get it~

pkg/deps/maven.go
```diff
@@ -0,0 +1,162 @@
+package deps
```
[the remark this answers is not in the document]
MoGuGuai-hzr: 😅

pkg/deps/maven_test.go
MoGuGuai-hzr: add option `-DincludeScope=runtime` to `mvn copy-dependencies` for excluding `test`, `provided` and `system`, and modify the test case

pkg/deps/maven.go
```diff
@@ -0,0 +1,233 @@
...
+	"github.com/apache/skywalking-eyes/license-eye/pkg/license"
+)
+
+var maven string
+
+func init() {
+	err := CheckMVN()
+	if err != nil {
+		logger.Log.Errorln("an error occurred when checking maven tool:", err)
+	}
```
kezhenxu94, line 43: Let's invoke this in maven resolver manually instead of in `init` method, so that when users don't need to resolve `pom.xml` files won't see these logs

MoGuGuai-hzr: have moved it~

kezhenxu94:
Screen Shot 2021-07-27 at 12 01 04

So many changed files 😱

MoGuGuai-hzr:
> 
> 
> Screen Shot 2021-07-27 at 12 01 04
> 
> So many changed files / lines 😱

It may be caused by the CR/LF, now it is fixed.

pkg/deps/maven.go
```diff
@@ -0,0 +1,229 @@
...
+	if err := os.Chdir(dir); err != nil {
+		logger.Log.Errorf("an error occurred when entering dir <%s> to parser file <%s>:%v\n", dir, base, err)
+		return false
+	}
+
+	err := resolver.CheckMVN()
+	if err != nil {
+		logger.Log.Errorln("an error occurred when checking maven tool:", err)
+		return false
+	}
```
kezhenxu94, line 92: Please bear in mind, `CanResolve` will be invoked for every kind of projects, not just maven projects, so if I just want to resolve Go project or NPM project, this error logs look weird to me. Note that I'm not saying we should not log this error, instead, we should only check `mvn` right before when we need to run `mvn` command, (in the `Resolve` method), `CanResolve` is only used to filter the resolver according to the files.

pkg/deps/maven.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+// CanResolve determine whether the file can be resolve by name of the file
+func (resolver *MavenPomResolver) CanResolve(mavenPomFile string) bool {
+	// switch to the directory where the file is located for searching mvnw
+	dir, base := resolver.Split(mavenPomFile)
+	if err := os.Chdir(dir); err != nil {
+		logger.Log.Errorf("an error occurred when entering dir <%s> to parser file <%s>:%v\n", dir, base, err)
+		return false
+	}
```
kezhenxu94, line 86: I think this part should be moved into `CheckMVN` function?

pkg/deps/maven.go
```diff
@@ -0,0 +1,229 @@
...
+
+// CheckMVN use mvn by default, use mvn if mvnw is not found
+func (resolver *MavenPomResolver) CheckMVN() error {
+	var err error
+
+	logger.Log.Debugln("searching mvnw ...")
+	_, err = exec.Command("./mvnw", "--version").Output()
+	if err == nil {
+		resolver.maven = "./mvnw"
+		logger.Log.Debugln("use mvnw")
```
kezhenxu94, line 115: `Debugf("mvnw is found at %v, will use mvnw by default", fileLocation)`

pkg/deps/maven.go
MoGuGuai-hzr: get it~

pkg/deps/maven.go
MoGuGuai-hzr: ok~

pkg/deps/maven.go
MoGuGuai-hzr: does this mean to modify the log message?

kezhenxu94: > does this mean to modify the log message?

Uh yes, sorry, didn't make it clear 😢

MoGuGuai-hzr: it is my fault, I didn't understand it well😅

MoGuGuai-hzr:
added some patterns describing the possible locations of license, but does not work for some jar packages

MoGuGuai-hzr (commented):
should i crawl the original webpage or judge directly from the url to identify the license

kezhenxu94:
> should i crawl the original webpage or judge directly from the url to identify the license

We can consider this later, we don't need to do all in one PR (which is hard to review and cost too long to finish), we can deliver minimal yet complete features step by step, in this PR I think it's OK to only support the "standard" dependencies that have a `LICENSE` file in their `.jar`, so let's polish this part instead of keep adding new things in this PR

MoGuGuai-hzr:
> 
> 
> > should i crawl the original webpage or judge directly from the url to identify the license
> 
> We can consider this later, we don't need to do all in one PR (which is hard to review and cost too long to finish), we can deliver minimal yet complete features step by step, in this PR I think it's OK to only support the "standard" dependencies that have a `LICENSE` file in their `.jar`, so let's polish this part instead of keep adding new things in this PR

ok~, i will do it

pkg/deps/maven.go
```diff
@@ -0,0 +1,591 @@
...
+
+// CheckMVN check available maven tools, find local repositories and download all dependencies
+func (resolver *MavenPomResolver) CheckMVN() error {
+	if err := resolver.FindMaven("./mvnw"); err == nil {
+		logger.Log.Debugln("mvnw is found, will use mvnw by default")
+	} else if err := resolver.FindMaven("mvn"); err != nil {
+		return fmt.Errorf("neither found mvnw nor mvn")
+	}
+
+	logger.Log.Debugln("mvnw is not found, will use mvn by default")
```
kezhenxu94, line 82: This is not needed / correct now,

pkg/deps/maven.go
```diff
@@ -0,0 +1,591 @@
...
+	if err != nil {
+		return err
+	}
+
+	resolver.repo = string(output)
+	return nil
+}
+
+func (resolver *MavenPomResolver) DownloadDeps() error {
+	cmd := exec.Command(resolver.maven, "process-resources") // #nosec G204
```
kezhenxu94, line 115: This seems to be problematic in multi-modules maven project, can you try this feature in our main repo http://github.com/apache/skywalking ?

pkg/deps/maven.go
MoGuGuai-hzr: ok, i have removed it

wu-sheng:
Just a reminder, I am not sure whether this could be fixed easily. Maven includes a plugin called `shaded`, which would not include the dependency jar, but package it inside another package during the `install`. You could try this case.

pkg/deps/maven.go
MoGuGuai-hzr: as you said in discussion, the failure may be caused by submodules. so i resolve it using maven command `install -DskipTests` to download possible dependencies, and it may take some time.

MoGuGuai-hzr:
> 
> 
> Just a reminder, I am not sure whether this could be fixed easily. Maven includes a plugin called `shaded`, which would not include the dependency jar, but package it inside another package during the `install`. You could try this case.

thanks for your help, and i am using `install` command attempting to resolve the incompatibility

wu-sheng:
> > Just a reminder, I am not sure whether this could be fixed easily. Maven includes a plugin called `shaded`, which would not include the dependency jar, but package it inside another package during the `install`. You could try this case.
> 
> thanks for your help, and i am using `install` command attempting to resolve the incompatibility

My point is, the module using this plugin, may escape from your dependency check. You should check the case.

MoGuGuai-hzr:
> 
> 
> > > Just a reminder, I am not sure whether this could be fixed easily. Maven includes a plugin called `shaded`, which would not include the dependency jar, but package it inside another package during the `install`. You could try this case.
> > 
> > 
> > thanks for your help, and i am using `install` command attempting to resolve the incompatibility
> 
> My point is, the module using this plugin, may escape from your dependency check. You should check the case.

sorry, I didn’t understand it very well. does this mean giving up on resolving these module? since the shaded package will be packaged during the `install`, can it be solved by installing first?

wu-sheng:
> orry, I didn’t understand it very well. does this mean giving up on resolving these module? since the shaded package will be packaged during the `install`, can it be solved by installing first?

I don't know. This is only a reminder, I just recommend you to check.

MoGuGuai-hzr:
> 
> 
> > orry, I didn’t understand it very well. does this mean giving up on resolving these module? since the shaded package will be packaged during the `install`, can it be solved by installing first?
> 
> I don't know. This is only a reminder, I just recommend you to check.

ok, I've got the information. i will attemp it later

pkg/license/identifier.go
```diff
@@ -80,3 +80,10 @@ func Identify(pkgPath, content string) (string, error) {
 
 	return "", fmt.Errorf("cannot identify license content")
 }
+
+var seemLicense = regexp.MustCompile(`(?i)licen[sc]e|copyright|copying`)
+
+// Seem determine whether the content of the file may be a license file
+func Seem(content string) bool {
+	return seemLicense.MatchString(content)
+}
```
kezhenxu94, line 89: Can you reuse `possibleLicenseFileName` in file `golang.go`? it may not be a good idea to put `possibleLicenseFileName` in `identifier.go` but now let's unify the 2 places, you can move `possibleLicenseFileName` into `identifier.go` later.

pkg/deps/maven_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+			if skipped := len(report.Skipped); skipped > 0 {
+				pkgs := make([]string, skipped)
+				for i, s := range report.Skipped {
+					pkgs[i] = s.Dependency
+				}
+
+				t.Errorf(
+					"there are %v dependencies in total, failed to identify the licenses of following packages (%d):\n", test.cnt,
+					len(pkgs))
```
kezhenxu94, line 134: `t.Errorf` should ONLY be used when the error is unexpected.

In this test case, we expect that some of the dependencies are not identified, right?

Anyway we should make the CI status passed

kezhenxu94 (approved):
Nice work! This is closing to serve as the 1st version for maven projects, though we may have some corner cases to consider later. 👍🏻 

I left some comments, please take a look

pkg/license/identifier.go
MoGuGuai-hzr: > 
> 
> Can you reuse `possibleLicenseFileName` in file `golang.go`? it may not be a good idea to put `possibleLicenseFileName` in `identifier.go` but now let's unify the 2 places, you can move `possibleLicenseFileName` into `identifier.go` later.

the purpose of `Seem` function was to simply determine whether one file may be a license base on the input content, not the file name. i remove it from `identifier.go` now.

pkg/deps/maven_test.go
MoGuGuai-hzr: i get it

kezhenxu94 (approved):
Thank you @MoGuGuai-hzr for the great work and patience 🙇🏻

PR #52, opened 2021-08-07
Enhance NPM dependency resolver to resolve deprecated license styles

zooltd:
This patch enhances the NPM dependency resolver to resolve deprecated license styles, which are like
```
{
  "license" : {
    "type" : "ISC",
    "url" : "https://opensource.org/licenses/ISC"
  }
}

{
  "licenses" : [
    {
      "type": "MIT",
      "url": "https://www.opensource.org/licenses/mit-license.php"
    },
    {
      "type": "Apache-2.0",
      "url": "https://opensource.org/licenses/apache2.0.php"
    }
  ]
}
```

### DIFF
--- a/pkg/deps/npm.go
+++ b/pkg/deps/npm.go
@@ -224,7 +224,7 @@ func (resolver *NpmResolver) ResolvePkgFile(pkgFile string) (string, error) {
 		return lcs, nil
 	}
 
-	return "", fmt.Errorf("cannot parse the \"license\"/\"licenses\" field")
+	return "", fmt.Errorf(`cannot parse the "license"/"licenses" field`)
 }
 
 // ResolveLicenseField parses and validates the "license" field in package.json file
--- a/pkg/deps/npm.go
+++ b/pkg/deps/npm.go
@@ -249,19 +249,16 @@ func (resolver *NpmResolver) ResolveLicenseField(rawData []byte) (string, bool) 
 }
 
 // ResolveLicensesField parses and validates the "licenses" field in package.json file
-// Additionally, the output is converted into the SPDX license expression syntax version 2.0 string, like "(ISC OR GPL-3.0)"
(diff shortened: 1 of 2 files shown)


pkg/deps/npm.go
```diff
@@ -206,10 +215,53 @@ func (resolver *NpmResolver) ResolvePkgFile(pkgFile string) (string, error) {
...
+				return t, true
+			}
+		}
+	}
+	return "", false
+}
+
+// ResolveLicensesField parses and validates the "licenses" field in package.json file
+// Additionally, the output is converted into the SPDX license expression syntax version 2.0 string, like "(ISC OR GPL-3.0)"
+func (resolver *NpmResolver) ResolveLicensesField(licenses []Lcs) (string, bool) {
```
kezhenxu94, line 253: Let's remove `()` in the licenses and simply this

```go
	var lcs []string
	for _, l := range licenses {
		lcs = append(lcs, l.Type)
	}
	if len(lcs) == 0 {
		return "", false
	}
	return strings.Join(lcs, " OR "), true
```

pkg/deps/npm.go
```diff
@@ -206,10 +215,53 @@ func (resolver *NpmResolver) ResolvePkgFile(pkgFile string) (string, error) {
...
+
+	if lcs, ok := resolver.ResolveLicenseField(packageInfo.License); ok {
+		return lcs, nil
+	}
+
+	if lcs, ok := resolver.ResolveLicensesField(packageInfo.Licenses); ok {
+		return lcs, nil
+	}
+
+	return "", fmt.Errorf("cannot parse the \"license\"/\"licenses\" field")
```
kezhenxu94, line 227:
```suggestion
	return "", fmt.Errorf(`cannot parse the "license"/"licenses" field`)
```

kezhenxu94 (approved):
Thanks

PR #53, opened 2021-08-11
add support for resolving jars' licenses

MoGuGuai-hzr:


wu-sheng:
Could you be clear in the PR comments? What is your resolving way?

MoGuGuai-hzr:
In fact, this is not one new implement. In the last pr, we searched for the license files that the maven project depends on from two places:

- searching in jar file for license from license[.txt] copying, manifest file and so on;
- searching in AID-Version.pom file for license from comments the head of the file and license element in xml

Now we separate jar file resolver from maven resolver to support that resolves a series of jar files entered by the user

wu-sheng:
We need a document to list the license resolving mechanism. Not just for Maven, for all supported language ecosystems.
This would be very helpful for users when they face license check fails. Such as having wrong mixed licenses, wrong in higher priority place, but a right and compatible one in lower priority place.

FYI @kezhenxu94

wu-sheng:
Let's create an issue to track this document requirements.

pkg/deps/jar_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+	if err := os.Chdir(dir); err != nil {
+		return nil, err
+	}
+
+	if err := dumpPomFile(pomFile, content); err != nil {
+		return nil, err
+	}
+
+	if _, err := exec.Command("mvn", "dependency:copy-dependencies", "-DoutputDirectory=./lib", "-DincludeScope=runtime").Output(); err != nil {
```
kezhenxu94, line 61: This is not portable, users / developers don't have `mvn` installed won't be able to run this test

pkg/deps/jar.go
```diff
@@ -0,0 +1,136 @@
...
+
+type JarResolver struct{}
+
+func (resolver *JarResolver) CanResolve(jarFile string) bool {
+	return filepath.Ext(jarFile) == ".jar"
+}
+
+func (resolver *JarResolver) Resolve(jarFile string, report *Report) error {
+	state := NotFound
+	if err := resolver.ResolveJar(&state, jarFile, report); err != nil {
```
kezhenxu94, line 41: In this resolver (jar resolver), we should support a path pattern like `dist/**/*.jar` and expand all the `.jar` files and resolve their licenses, the current feature in this PR only support a singe `.jar` file, right?

kezhenxu94 (commented):
This feature is for a scenario where users already have their distribution packages and check the final `.jar` files, the key point is that we need to expand all `.jar` files in the patter like:

```yaml
dependency:
  files:
    - go.mod
    - dist/**/*.jar
```

pkg/deps/jar_test.go
MoGuGuai-hzr: The purpose of this test is to check the usability of the JarResolver. Can I put the jar files directly in the test/testdata?

pkg/deps/jar.go
MoGuGuai-hzr: From my understanding, JarResolver is only responsible for resolve the licenses of jar files. And `deps/config.go` will enter all jar files in the `dist/**/*.jar`

pkg/deps/jar_test.go
kezhenxu94: > The purpose of this test is to check the usability of the JarResolver. Can I put the jar files directly in the test/testdata?

you can put binary files (`.jar`) files in the codebase but we can't release them, this is unusual case, maybe we can only run this test case when `mvn` is available, otherwise we skip this test?

pkg/deps/jar.go
kezhenxu94: > From my understanding, JarResolver is only responsible for resolve the licenses of jar files. And `deps/config.go` will enter all jar files in the `dist/**/*.jar`

Well, now, it seems `deps/config.go` doesn't expand the pattern `dist/**/*.jar` to all `.jar` files (like `dist/module1/log4j.jar`, `dist/module2/junit.jar`), we should do this, right?

pkg/deps/jar_test.go
MoGuGuai-hzr: > 
> you can put binary files (`.jar`) files in the codebase but we can't release them, this is unusual case, maybe we can only run this test case when `mvn` is available, otherwise we skip this test?

ok, i will check the availability of mvn first

pkg/deps/jar.go
MoGuGuai-hzr: >
> Well, now, it seems `deps/config.go` doesn't expand the pattern `dist/**/*.jar` to all `.jar` files (like `dist/module1/log4j.jar`, `dist/module2/junit.jar`), we should do this, right?

Yes, it will be complete in the next pr

kezhenxu94: > > Well, now, it seems `deps/config.go` doesn't expand the pattern `dist/**/*.jar` to all `.jar` files (like `dist/module1/log4j.jar`, `dist/module2/junit.jar`), we should do this, right?
> 
> Yes, it will be complete in the next pr

Nice

kezhenxu94 (approved):

wu-sheng:
How far are we from using this in main repo and the next release?

kezhenxu94:
> How far are we from using this in main repo and the next release?

For now this feature is only helpful to assist human to audit the licenses, the key to make it available in main repo (automatically report conflict licenses and fail the CI) is to provide a license compatibility matrix and check the final resolved licenses. We might provide another command `license-eye dependency check`, (this command is `license-eye dependency resolve`)

PR #54, opened 2021-08-13
set log level to info in GitHub Action usage

fgksgf:


kezhenxu94 (approved):

tisonkun:
I'd prefer we can use an option to config it.

kezhenxu94:
> I'd prefer we can use an option to config it.

Reasonable.

PR #57, opened 2021-08-16
make the config file configurable

fgksgf:


kezhenxu94 (approved):
Thanks!! 🙇🏻

wu-sheng:
-c seems not in the doc.

wu-sheng:
https://github.com/apache/skywalking-eyes#configurations This should be updated too.

kezhenxu94:
> https://github.com/apache/skywalking-eyes#configurations This should be updated too.

The `configurations` section is all about the **contents** in config file `.licneserc.yaml`, and `-c` is a pointer to the config file itself.

> -c seems not in the doc.

It's in the doc, in example command
Screen Shot 2021-08-16 at 15 05 40

PR #58, opened 2021-08-18
Set up mechanism to check dependencies' license compatibilities

kezhenxu94:
The compatible matrix needs further polishes, but his sets up the mechanism.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,8 @@ INFO Totally checked 20 files, valid: 10, invalid: 10, ignored: 0, fixed: 10
 
 #### Resolve Dependencies' licenses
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -135,6 +137,33 @@ ERROR failed to identify the licenses of following packages:
 gopkg.in/yaml.v3
 ```
 
+#### Check Dependencies' licenses
```
wu-sheng, line 140: Should we provide PR comment after the check?

wu-sheng (approved):
Cool!

wu-sheng:
Let's run for the source code itself?

kezhenxu94:
> Let's run for the source code itself?

Some dependencies' licenses are still unrecognized, I need to polish and resolve them first.

PR #65, opened 2021-09-03
Feature: `header fix` supports nextflow files

keiranmraine:
Tested and shown to work with no impact on files.

Nextflow uses Java/Groovy comments.


kezhenxu94 (approved):
Thank you @keiranmraine

PR #67, opened 2021-09-08
Resolve absolute path in `.gitignore` to relative path

kezhenxu94:
Closes https://github.com/apache/skywalking/issues/7543

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,7 @@
 # under the License.
 # 
 .idea/
-bin/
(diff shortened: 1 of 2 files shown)


kezhenxu94:
Trimming the leading `/` should fix the bug, `whatever/whatever/bin/whatever.go` will NOT be ignored, this is the same behavior as `.gitignore`

wu-sheng (approved):

PR #69, opened 2021-09-10
feature: support saving dependencies' licenses

kezhenxu94:
This is a missing but important feature that helps to organize the distribution package especially in ASF projects, it resolves and saves all LICENSE files of the dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,8 +112,11 @@ INFO Totally checked 20 files, valid: 10, invalid: 10, ignored: 0, fixed: 10
 
 This command serves as assistance for human beings to audit the dependencies license, it's exit code is always 0.
 
+You can also use the `--output` or `-o` to save the dependencies' `LICENSE` files to a specified directory so that
+you can put them in distribution package if needed.
+
 ```bash
-bin/darwin/license-eye -c test/testdata/.licenserc_for_test_check.yaml dep resolve
+bin/darwin/license-eye -c test/testdata/.licenserc_for_test_check.yaml dep resolve -o ./dependencies/licenses
 INFO GITHUB_TOKEN is not set, license-eye won't comment on the pull request
 INFO Loading configuration from file: test/testdata/.licenserc_for_test_check.yaml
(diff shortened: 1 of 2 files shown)


wu-sheng (approved):
This feature is good to me. From ASF  perspective, not all license files should be copied.

zooltd:
This feature attempts to write the each pkg's license content to respective files. The license content is resolved by each resolver. However, the resolving behavior of each resolver is inconsistent. To be more specific, in `GoModResolver`, it always record a mod's complete license content, cuz it identifies the mod's license through this content. In `NpmResolver`, it may not record the license content because it identifies a pkg's license through the field in a pkg's `package.json` file in the first priority. In `MavenPomResolver`, it simply records the raw data of the license field in a `pom.xml` file as the license content. So here's the problem, the license content to be written may be an empty string or something irrelevant. Maybe we should standardize each resolver's behavior. If the `LicenseContent` field to be written is empty, then simply write a license template to the file according to the `LicenseSpdxID` field.

kezhenxu94:
> This feature attempts to write each pkg's license content to respective files. The license content is resolved by each resolver. However, the resolving behavior of each resolver is inconsistent. To be more specific, in `GoModResolver`, it always record a mod's complete license content, cuz it identifies the mod's license through this content. In `NpmResolver`, it may not record the license content because it identifies a pkg's license through the field in a pkg's `package.json` file in the first priority. In `MavenPomResolver`, it simply records the raw data of the license field in a `pom.xml` file as the license content. So here's the problem, the license content to be written may be an empty string or something irrelevant. Maybe we should standardize each resolver's behavior. If the `LicenseContent` field to be written is empty, then simply write a license template to the file according to the `LicenseSpdxID` field.

Yes there are still some resolvers that don't record the content, some of them record a spdx id and some of them record the license URL, we should make sure the Content field really saves the content, we may add other Fields that can be empty, like URL...

kezhenxu94:
@zooltd will you be available to enhance this part?

zooltd:
> @zooltd will you be available to enhance this part?

Yep, glad to help.😊

PR #70, opened 2021-09-15
fix the save location problem

zooltd:
During resolving licenses, the working directory of the process would change. In this situation, if the `output` location a user inputed is a relative path, then the saving location would be different from what it was defined at first.


kezhenxu94 (approved):
thank you!!! 🙇🏻

PR #71, opened 2021-09-15
Enhance Npm Resolver: support recording license contents

zooltd:
This patch eanhances Npm resolver to record each pkg's license contents. In previous versions, Npm resolver will not record the license content when it identifies a pkg's license through the package.json file. Now, it will still resolve a pkg's license info through its package.json file first. Also, whether or not the previous step succeeded, it will try to locate a pkg's LICENSE file and record its content.

### DIFF
--- a/pkg/deps/result.go
+++ b/pkg/deps/result.go
@@ -31,12 +31,11 @@ const (
 
 // Result is a single item that represents a resolved dependency license.
 type Result struct {
-	Dependency       string
-	LicenseFilePath  string
-	LicenseContent   string
-	LicenseSpdxID    string
-	ManifestFilePath string
-	ResolveErrors    []error
+	Dependency      string
+	LicenseFilePath string
+	LicenseContent  string
+	LicenseSpdxID   string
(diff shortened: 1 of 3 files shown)


pkg/deps/result.go
```diff
@@ -31,10 +31,12 @@ const (
...
 type Result struct {
-	Dependency      string
-	LicenseFilePath string
-	LicenseContent  string
-	LicenseSpdxID   string
+	Dependency       string
+	LicenseFilePath  string
+	LicenseContent   string
+	LicenseSpdxID    string
+	ManifestFilePath string
```
kezhenxu94, line 38: This seems to be unused now, do you think it will be used any soon?

kezhenxu94 (approved):
LGTM, one nit

pkg/deps/result.go
zooltd: Nope. The `ManifestFilePath` field seems redundent. I'm gonna delete this field.

kezhenxu94 (approved):

PR #74, opened 2021-09-29
enhance MavenPomResolver's capabilities

MoGuGuai-hzr:
Using `mvn clean install` will take a lot of time to compile, which is not what we expect.

Therefore, we add a new dependency resolution: splitting dependencies in each `pom.xml` to make them independent. Of course, each module will inherit the corresponding dependencies from the parent module.


kezhenxu94:
Hi @MoGuGuai-hzr , ping me when it's ready to review, or you need any help 😄

MoGuGuai-hzr:
> 
> 
> Hi @MoGuGuai-hzr , ping me when it's ready to review, or you need any help 😄

The currently submitted version already support to resolve the dependencies of project skywalkin, but it takes too much time.

I'm trying to optimize it

wu-sheng:
No update for months. Closing for now.

PR #76, opened 2021-10-02
Support installing via go install

kezhenxu94:
For reviewers who want to try, please use `go install github.com/apache/skywalking-eyes/cmd/license-eye@goinstall`


wu-sheng (commented):
A question, what goinstall could provide?

kezhenxu94:
> A question, what goinstall could provide?

It's a more Go-native way to install Go tools, more convenient for demonstration, less steps (only one command) to quick start.

kezhenxu94:
Consider it as the `brew` command in Go ecosystem.

wu-sheng:
> Consider it as the `brew` command in Go ecosystem.

Love brew 😄

wu-sheng (approved):

PR #77, opened 2021-10-08
Added .pre-commit-hooks.yaml

HungryJoe:
This addition allows users of the [pre-commit](https://pre-commit.com/) Git hook management tool to use Skywalking Eyes' `license-eyes header fix` command to automate fixing their license headers at regular points in the development lifecycle.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,5 @@
+-   id: fix-license
+    name: fix-license
+    entry: license-eye header fix
+    language: golang
+    description: Add license headers to files missing them


kezhenxu94:
[pre-commit](https://pre-commit.com/) requires third-party dependency to be installed in developers machine, which I don't think is a good idea, can we just use the native pre-commit hook provided by git? @HungryJoe , also, this file only takes effects in this repo, I'd rather we provide a new command like `skywalking-eyes header git-hook` to automatically create the needed files into the git codebase.

kezhenxu94:
@HungryJoe any update?

PR #84, opened 2021-11-02
Adds an option to the action to run in "fix" mode

ryanmrichard:
This PR is a small change which adds an additional, optional, input to the action called `mode`. My goal with this change is to make it so CI pipelines can automatically apply the licenses to PRs which lack them (applying the changes would be handled by a different action like [this one](https://github.com/EndBug/add-and-commit)). I looked around and did not see any mention of such functionality; apologies if I missed it.

Some additional notes:

- `mode` can take one of two values: `check` or `fix`. I set the default to `check` to preserve existing behavior.
- I have also updated the README with the new input description. 
- I see from the diff that my IDE removed some trailing spaces; if you really want those back I can redo the changes with vim or something
- I have tested that this works as intended with one of my own repositories; if you want proper unit tests I will need guidance/help on how to go about doing that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ To use License-Eye in GitHub Actions, add a step in your GitHub workflow.
       # log: debug # optional: set the log level. The default value is `info`.
       # config: .licenserc.yaml # optional: set the config file. The default value is `.licenserc.yaml`.
       # token: # optional: the token that license eye uses when it needs to comment on the pull request. Set to empty ("") to disable commenting on pull request. The default value is ${{ github.token }}
+      # mode: # optional: Which mode License Eye should be run in. Choices are `check` or `fix`. The default value is `check`.
 ```
 
 Add a `.licenserc.yaml` in the root of your project, for Apache Software Foundation projects, the following configuration should be enough.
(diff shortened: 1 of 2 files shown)


kezhenxu94 (commented):
@ryanmrichard thank you very much for the patch.

can you please also add some doc to mention that users need to use other GitHub Action to commit the fixed changes and the exit code of fix mode is always 0 so they cannot rely on the exit code to block the CI status?

ryanmrichard:
@kezhenxu94 I added a subsection detailing fix mode usage per your request.

README.md
```diff
@@ -44,6 +44,33 @@ header:
 
 **NOTE**: The full configurations can be found in [the configuration section](#configurations).
 
+#### Using the Action in Fix Mode
+
+By default the action runs License-Eye in check mode, which will raise an error
+if any of the processed files are missing license headers. If `mode` is set to
+`fix`, the action will instead apply the license header to any processed file
+that is missing a license header. The fixed files can then be pushed back to the
+pull request using another GitHub action. For example:
```
wu-sheng, line 53: So, one action would commit and push, then make the CI failing, right? Just confirming. Then maintainers should rerun the CI task to make it passed.

ryanmrichard: @wu-sheng The SkyWalking-Eyes action will edit the files in place. The subsequent "add-and-commit" action will use git to commit the changed files back to the PR. The action I linked to has a couple returns (see [here](https://github.com/EndBug/add-and-commit#outputs)); if you wanted to, you could add a step to the action which errs if the "add-and-commit" action makes a commit. Automating the re-run is not straightforward because GitHub doesn't like workflows triggering other workflows, so the maintainer will probably need to manually re-run the failed workflow. 

In practice I usually don't worry about this becasue there's usually other changes needed to the PR; so when the PR author submits those changes (which will necessarily need to be synched with the automatic commit) they'll re-trigger CI.

PR #89, opened 2022-02-04
Add comment style for HCL (Terraform)

stumins:
Add comment style for HCL (Terraform) to enable `header fix` support

https://www.terraform.io/language/syntax/configuration#comments


wu-sheng:
Looks good. Wait for @kezhenxu94 confirmatiom

kezhenxu94 (approved):
Thanks

PR #91, opened 2022-02-12
config: Fix incorrect log message

Xuanwo:


pkg/config/config.go
```diff
@@ -45,7 +45,7 @@ func (config *Config) Parse(file string) (err error) {
 	}
 
 	if os.IsNotExist(err) {
-		logger.Log.Infof("No config file is given, using the default config")
+		logger.Log.Infof("Config file %s is not exist, using the default config", file)
```
wu-sheng, line 48:
```suggestion
		logger.Log.Infof("Config file %s does not exist, using the default config", file)
```

English grammar issue.

kezhenxu94 (approved):

PR #93, opened 2022-02-19
Fix license check in go library testify

mrproliu:
I use the `logrus` library in the go and it's dependent on the old version of the testify library. After I run `dep check`, the license of this library could not analyze.

Here is the `LICENSE` file link: https://raw.githubusercontent.com/stretchr/testify/v1.2.2/LICENSE

### DIFF
--- a/pkg/license/norm.go
+++ b/pkg/license/norm.go
@@ -255,6 +255,11 @@ var (
 			regexp.MustCompile(`(?im)^\s*.+ is distributed under the Simplified BSD License\:?$`),
 			"",
 		},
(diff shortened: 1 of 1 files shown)


wu-sheng (commented):
Is this a BSD

mrproliu:
Yes. I think It's a note added by the Contributor and could be ignored to complete the license check.

wu-sheng (approved):
Wait for @kezhenxu94

kezhenxu94 (approved):

PR #94, opened 2022-03-11
Trim leading and trailing newlines before rewrite license header cotent

tisonkun:
This closes https://github.com/apache/skywalking/issues/8529.

Also, rewrite all our files after the fix and a new content aligned template.

This is related to https://github.com/apache/skywalking/discussions/8657.

cc @wu-sheng @kezhenxu94


pkg/header/fix.go
```diff
@@ -101,7 +100,10 @@ func GenerateLicenseHeader(style *comments.CommentStyle, config *ConfigHeader) (
 		return "", err
 	}
 
-	lines := strings.Split(config.GetLicenseContent(), "\n")
+	content := config.GetLicenseContent()
+	// trailing leading and trailing newlines
```
kezhenxu94, line 104: Do you mean this?

```suggestion
	// Trim leading and trailing newlines
```

kezhenxu94 (approved):
LGTM

pkg/header/fix.go
tisonkun: Yes. Thanks for your suggestion.

PR #103, opened 2022-05-06
Provide `--summary` flag to generate the license summary file

mrproliu:
Follow https://github.com/apache/skywalking/discussions/8992, I have done two things:
1. Add a new flag `--summary` to help users generate the license summary file by template file. 
2. Add `licenses` config under the `dependency` to declare which license is could not be recognized.

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,6 +363,298 @@ gopkg.in/check.v1
 
 </details>
 
+##### Summary Template
+
+The summary is a template to generate the summary of dependencies' licenses based on the [Golang Template](https://pkg.go.dev/text/template). It includes these variables:
+
+|Name|Type|Example|Description|
+|----|----|-------|-----------|
+|LicenseContent|string|`{{.LicenseContent}}`|The project license content, it's relate to the `header.license.spdx-id` config. |
+|Groups|list structure|`{{ range .Groups }}`|The dependency group, all license is aggregate by the same license [SPDX ID](https://spdx.org/licenses/). |
+|Groups.Name|string|`{{.Name}}`|The [SPDX ID](https://spdx.org/licenses/) of dependency. |
+|Groups.Deps|list structure|`{{ range .Deps }}`|All dependencies, it including all of same [SPDX ID](https://spdx.org/licenses/) dependencies under `Groups.Name`. |
+|Groups.Deps.Name|string|`{{.Name}}`|The name of the dependency. |
+|Groups.Deps.Version|string|`{{.Version}}`|The version of the dependency. |
+|Groups.Deps.LicenseID|string|`{{.LicenseID}}`|The [SPDX ID](https://spdx.org/licenses/) of the dependency license. |
+
+<details>
+<summary>Summary Template Generate</summary>
+
+Summary template content:
+```
+{{.LicenseContent }}
+{{ range .Groups }}
+========================================================================
+{{.Name}} licenses
+========================================================================
+{{range .Deps}}
+    {{.Name}} {{.Version}} {{.LicenseID}}
+{{- end }}
+{{ end }}
+```
+
+Generate LICENSE file content:
+```
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
+      "Contribution" shall mean any work of authorship, including
+      the original version of the Work and any modifications or additions
+      to that Work or Derivative Works thereof, that is intentionally
+      submitted to Licensor for inclusion in the Work by the copyright owner
+      or by an individual or Legal Entity authorized to submit on behalf of
+      the copyright owner. For the purposes of this definition, "submitted"
+      means any form of electronic, verbal, or written communication sent
+      to the Licensor or its representatives, including but not limited to
+      communication on electronic mailing lists, source code control systems,
+      and issue tracking systems that are managed by, or on behalf of, the
+      Licensor for the purpose of discussing and improving the Work, but
+      excluding communication that is conspicuously marked or otherwise
+      designated in writing by the copyright owner as "Not a Contribution."
+
+      "Contributor" shall mean Licensor and any individual or Legal Entity
+      on behalf of whom a Contribution has been received by Licensor and
+      subsequently incorporated within the Work.
+
+   2. Grant of Copyright License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      copyright license to reproduce, prepare Derivative Works of,
+      publicly display, publicly perform, sublicense, and distribute the
+      Work and such Derivative Works in Source or Object form.
+
+   3. Grant of Patent License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      (except as stated in this section) patent license to make, have made,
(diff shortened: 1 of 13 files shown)


commands/deps_resolve.go
```diff
@@ -102,3 +116,21 @@ func writeLicense(result *deps.Result) {
...
+	}
+	defer file.Close()
+	summary, err := deps.GenerateSummary(summaryTpl, &Config.Header, rep)
+	if err != nil {
+		logger.Log.Errorf("failed to generate summary content: %v", err)
+		return
+	}
+	_, err = file.WriteString(summary)
+	if err != nil {
+		logger.Log.Errorf("failed to write summary file, %v", err)
```
kezhenxu94, line 134: Return the error and handle it from the caller method

pkg/deps/golang.go
```diff
@@ -119,7 +125,10 @@ func (resolver *GoModResolver) ResolvePackageLicense(module *packages.Module, re
 			}
 			identifier, err := license.Identify(module.Path, string(content))
 			if err != nil {
-				return err
+				if declareLicense == nil {
```
kezhenxu94, line 128: If the users already declare this license I don't think we need to `license.Identify` it, this gives the users ability to override the license if we make something wrong

pkg/deps/summary.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+type SummaryRenderLicense struct {
+	Name      string // Dependency name
+	Version   string // Dependency version
+	Location  string // The filename of generated license file
+	LicenseID string // License ID
+}
+
+func GenerateDependencyLicenseFilename(result *Result) string {
```
kezhenxu94, line 50: This method is not used in `--summary` flag, it's used in `--output` flag, why put it here (`summary.go`)?

pkg/deps/summary.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	return r.String(), nil
+}
+
+func generateSummaryRenderContext(head *header.ConfigHeader, rep *Report) (*SummaryRenderContext, error) {
+	// the license id of the project
+	licenseContent, err := license.GetLicenseContent(head.License.SpdxID)
+	if err != nil {
+		return nil, err
+	}
```
kezhenxu94, line 85: `head.License.SpdxID` is not required so it might be empty, users can use other forms to specify their license, not just `SpdxID`. You need to take care the case where `SpdxID` is not set

pkg/deps/summary.go
mrproliu: Because I provide the `Location` field [into the template](https://github.com/apache/skywalking-eyes/pull/103/files#diff-d235c0f816b8e113b0fe5f573372f3bb09ed71a44c553056e452c5a9b5e59dc5R101), It could help for getting generated license file path.

pkg/deps/summary.go
kezhenxu94: > Sure, If the `head.License.SpdxID` is not set, should I just throw a new error? or another config could get the license?

You can use this method to get the license content, but notice only use this as a fallback when the users don't specify spdxID https://github.com/apache/skywalking-eyes/blob/985866ce7e324454f61e22eb2db2e998db09d6f3/pkg/header/config.go#L171

pkg/deps/golang.go
mrproliu: Updated.

pkg/deps/summary.go
mrproliu: Updated.

commands/deps_resolve.go
```diff
@@ -38,9 +39,11 @@ func init() {
 	DepsResolveCommand.PersistentFlags().StringVarP(&outDir, "output", "o", "",
 		"the directory to output the resolved dependencies' licenses, if not set the dependencies' licenses won't be saved")
 	DepsResolveCommand.PersistentFlags().StringVarP(&summaryTplPath, "summary", "s", "",
-		"the template file to write the summary the dependencies' licenses, if not set the licenses just print as the table")
+		"the template file to write the summary the dependencies' licenses into \"LICENSE\" file into directory which same with template file.")
```
kezhenxu94, line 42:
```suggestion
		"the template file to write the summary of dependencies' licenses, a new file named \"LICENSE\" will be created in the same directory as the template file, to save the final summary.")
```

commands/deps_resolve.go
```diff
@@ -57,9 +60,11 @@ var DepsResolveCommand = &cobra.Command{
 			}
 		}
 		if summaryTplPath != "" {
-			if outDir == "" {
-				return fmt.Errorf("please provide the output directory to write the license summary file")
```
kezhenxu94: I think you need to keep the logic of `outDir`?

commands/deps_resolve.go
```diff
@@ -70,6 +86,12 @@ var DepsResolveCommand = &cobra.Command{
 			}
 		}
 
+		if summaryTpl != nil {
+			if err := writeSummary(&report); err != nil {
+				logger.Log.Warnf("write summary file error: %v", err)
```
kezhenxu94, line 91: What about

```suggestion
				return err
```

pkg/license/identifier.go
```diff
@@ -102,3 +102,25 @@ func Identify(pkgPath, content string) (string, error) {
 		return "", fmt.Errorf("cannot identify license content")
 	}
 }
+
+// GetLicenseContent from license id
+func GetLicenseContent(spdxID string) (string, error) {
+	for _, dir := range templatesDirs {
```
kezhenxu94, line 108: `templatesDirs` includes `lcs-templates` and `header-templates`, I think in this case we only need to walk through `lcs-templates`

README.md
```diff
@@ -157,11 +157,15 @@ INFO Totally checked 20 files, valid: 10, invalid: 10, ignored: 0, fixed: 10
...
+We also support two flags:
+
+|Flag name|Short name|Description|
+|---------|----------|-----------|
+|`--output`|`-o`|Save the dependencies' `LICENSE` files to a specified directory so that you can put them in distribution package if needed.|
+|`--summary`|`-s`|Based on the template, aggregate all dependency information and generate a `LICENSE` file.|
 
 ```bash
-license-eye -c test/testdata/.licenserc_for_test_check.yaml dep resolve -o ./dependencies/licenses
+license-eye -c test/testdata/.licenserc_for_test_check.yaml dep resolve -o ./dependencies/licenses -s LICENSE.tpl
```
kezhenxu94, line 168: Can you also add an section to demonstrate how the `LICENSE.tpl` looks like and what variables it can use?

commands/deps_resolve.go
mrproliu: I think it's better to perform before writing licenses if we need to throw the exception, it could make the command more consistent.

README.md
mrproliu: Updated.

README.md
```diff
@@ -359,6 +363,298 @@ gopkg.in/check.v1
 
 </details>
 
+##### Summary Template
+
+The summary is a template to generate the summary of dependencies' licenses based on the [Golang Template](https://pkg.go.dev/text/template). It includes these variables:
+
+|Name|Type|Example|Description|
+|----|----|-------|-----------|
+|LicenseContent|string|`{{.LicenseContent}}`|The project license content, it's relate to the `header.license.spdx-id` config. |
```
kezhenxu94, line 372:
```suggestion
|LicenseContent|string|`{{.LicenseContent}}`|The project license content, it's the license of `header.license.spdx-id` (if set), otherwise it's the `header.license.content`. |
```

README.md
```diff
@@ -359,6 +363,298 @@ gopkg.in/check.v1
...
+|Groups|list structure|`{{ range .Groups }}`|The dependency group, all license is aggregate by the same license [SPDX ID](https://spdx.org/licenses/). |
```
kezhenxu94, line 373:
```suggestion
|Groups|list structure|`{{ range .Groups }}`|The dependency groups, all licenses are grouped by the same license [SPDX ID](https://spdx.org/licenses/). |
```

README.md
```diff
@@ -359,6 +363,298 @@ gopkg.in/check.v1
...
+|Groups.Name|string|`{{.Name}}`|The [SPDX ID](https://spdx.org/licenses/) of dependency. |
+|Groups.Deps|list structure|`{{ range .Deps }}`|All dependencies, it including all of same [SPDX ID](https://spdx.org/licenses/) dependencies under `Groups.Name`. |
```
kezhenxu94, line 375:
```suggestion
|Groups.Deps|list structure|`{{ range .Deps }}`|All dependencies with the same [SPDX ID](https://spdx.org/licenses/). |
```

README.md
```diff
@@ -359,6 +363,298 @@ gopkg.in/check.v1
...
+
+<details>
+<summary>Summary Template Generate</summary>
+
+Summary template content:
+```
+{{.LicenseContent }}
+{{ range .Groups }}
+========================================================================
+{{.Name}} licenses
```
kezhenxu94, line 388: Do you think it'd be more specific if we rename `{{.Name}}` to something like `{{.LicenseSpdxID}}` or `{{.LicenseName}}`? Because there is another `{{.Name}}` inside the `{{.Deps}}`,

pkg/license/identifier.go
```diff
@@ -102,3 +104,23 @@ func Identify(pkgPath, content string) (string, error) {
...
+			continue
+		}
+
+		res, err := assets.Asset(filepath.Join(licenseTemplatesDir, f.Name()))
+		if err != nil {
+			return "", err
+		}
+		return string(res), nil
+	}
+	return "", fmt.Errorf("could not found license: %s", spdxID)
```
kezhenxu94, line 125: Can we just join the `licenseTemplateDir`, `spdxID` and `.txt` to avoid iterating all the files over and over again?

```suggestion
	res, err := assets.Asset(filepath.Join(licenseTemplatesDir, spdxID, ".txt"))
	if err != nil {
		return "", err
	}
	return string(res), nil
```

pkg/deps/summary.go
```diff
@@ -0,0 +1,99 @@
...
+		return "", err
+	}
+	return r.String(), nil
+}
+
+func generateSummaryRenderContext(head *header.ConfigHeader, rep *Report) *SummaryRenderContext {
+	// the license id of the project
+	var headerContent string
+	if head.License.SpdxID != "" {
+		headerContent, _ = license.GetLicenseContent(head.License.SpdxID)
```
kezhenxu94, line 67: I don't think the error can be ignored, if the users specify an unknown `SpdxID` (by mistake), we should notify that

kezhenxu94 (commented):
Some nits, generally good to me, thanks!

README.md
mrproliu: I think it could be better to rename it to `{{.LicenseID}}`? WDYT? 
Let it same with the `{{.LicenseID}}` inside the `{{.Deps}}`. For now, they are the same.

kezhenxu94: > I think it could be better to rename it to `{{.LicenseID}}`? WDYT?

`{{.LicenseID}}` is good to me

mrproliu: Updated.

pkg/deps/summary.go
mrproliu: Updated. If `GetLicenseContent` could not found, then throw a error.

kezhenxu94 (approved):

PR #104, opened 2022-05-10
Add GitHub Action step summary for better experience

kezhenxu94:
Now we have a summary just in the summary page, no need to click into the logs and check the logs

Screen Shot 2022-05-10 at 15 52 56


wu-sheng (approved):
👍

PR #105, opened 2022-05-10
Release 0.3.0

fgksgf:


kezhenxu94 (dismissed):

wu-sheng:
PlantUML SVG Diagrams CI fails?

kezhenxu94:
Well, plantUML check failed, let's just check the `.svg` into the codebase, I'll migrate PlantUML to mermaid diagram next

fgksgf:
It seems that PlantUML will generate different svg files on different OS though we do not change the source file. I think we can just keep svg and plantuml files but remove the check. Because they will not be updated frequently, right?

kezhenxu94:
> It seems that PlantUML will generate different svg files on different OS though we do not change the source file. I think we can just keep svg and plantuml files but remove the check. Because they will not be updated frequently, right?

ok

wu-sheng (approved):

PR #107, opened 2022-05-18
Use Google's licensecheck to identify licenses

fgksgf:
- Use [Google's licensecheck](https://github.com/google/licensecheck) to identify licenses, which is used by pkg.go.dev
- Add binary licenses by `dep resolve` (dogfooding)


fgksgf:
When execute `./bin/darwin/license-eye dep resolve -o dist/licenses -s dist/LICENSE.tpl`, got:

```
Failed to resolve the license of <github.com/kr/logfmt@v0.0.0-20140226030751-b84e30acd515>: cannot find license file 
Failed to resolve the license of <github.com/pascaldekloe/goe@v0.0.0-20180627143212-57f6aae5913c>: cannot identify the license, coverage: 21.6% 
```

Then I explored their repositories and found that they add their licenses recently, but the versions that we depend on do not have valid licenses. 

So do we need add their current licenses or not? @kezhenxu94

kezhenxu94:
> When execute `./bin/darwin/license-eye dep resolve -o dist/licenses -s dist/LICENSE.tpl`, got:
> 
> 
> 
> ```
> 
> Failed to resolve the license of <github.com/kr/logfmt@v0.0.0-20140226030751-b84e30acd515>: cannot find license file 
> 
> Failed to resolve the license of <github.com/pascaldekloe/goe@v0.0.0-20180627143212-57f6aae5913c>: cannot identify the license, coverage: 21.6% 
> 
> ```
> 
> 
> 
> Then I explored their repositories and found that they add their licenses recently, but the versions that we depend on do not have valid licenses. 
> 
> 
> 
> So do we need add their current licenses or not? @kezhenxu94 

Can we try to upgrade to the newer versions that have licenses? That would be more safe

fgksgf:
> Can we try to upgrade to the newer versions that have licenses? That would be more safe

Since we depend on them indirectly, I executed `go get -u ./... && go mod tidy` to upgrade all dependencies. Then we still have one dependency (indirect) that can not resolve the license:

```
Failed to resolve the license of <github.com/chzyer/logex@v1.1.10>: cannot find license file
```

kezhenxu94:
> > Can we try to upgrade to the newer versions that have licenses? That would be more safe
> 
> Since we depend on them indirectly, I executed `go get -u ./... && go mod tidy` to upgrade all dependencies. Then we still have one dependency (indirect) that can not resolve the license:
> 
> ```
> Failed to resolve the license of <github.com/chzyer/logex@v1.1.10>: cannot find license file
> ```

@fgksgf I just wanted to suggest you to add the following config into `.licenserc.yaml` but found it doesn't work for this dependency due tu #108 

```yaml
dependency:
  files: #...
  licenses:
    - name: github.com/chzyer/logex
      version: v1.1.10
      license: MIT
```

kezhenxu94:
For this kind of old dependencies, we need the `licenses` config to manually identify the licenses

pkg/license/identifier.go
```diff
@@ -19,58 +19,48 @@ package license
...
+	}
+)
 
-func init() {
-	wg := sync.WaitGroup{}
-	for _, dir := range templatesDirs {
-		files, err := assets.AssetDir(dir)
+// scanner returns a licensecheck.Scanner instance with its build-in licenses.
+// It will be initialized once.
+func scanner() *licensecheck.Scanner {
```
kezhenxu94, line 51: can we initialize `_scanner` in `init` method? And I think Go doesn't require `_` prefix to indicate private variable?

pkg/license/identifier.go
```diff
@@ -19,58 +19,48 @@ package license
...
-}
+var (
+	_scanner    *licensecheck.Scanner
+	scannerOnce sync.Once
 
-var normalizedTemplates = sync.Map{}
+	dualLicensePatterns = []*regexp.Regexp{
+		regexp.MustCompile(`(?i)This project is covered by two different licenses: (?P<license>[^.]+)`),
+	}
+)
```
kezhenxu94, line 47: Hey @fgksgf , I found `coverage.Match` is exactly what we want here, `coverage.Match` is an array of licenses if it's dual-license, just like what you put in the test, so we can simply remove this pattern and return the `coverage.Match` (for example, joined by `AND`)

kezhenxu94 (approved):
This solution is much more neat than before and we don't have to maintain the ugly rules... Thanks for doing this @fgksgf just some nits that you might want to improve

pkg/license/identifier.go
```diff
@@ -19,58 +19,48 @@ package license
...
+
 	"github.com/apache/skywalking-eyes/assets"
 	"github.com/apache/skywalking-eyes/internal/logger"
 )
 
-var licenseTemplatesDir = "lcs-templates"
+const (
+	// coverageThreshold is the minimum percentage of the file that must contain license text.
+	// Reference: https://github.com/golang/pkgsite/blob/d43359e3a135fc391960db4f5800eb081d658412/internal/licenses/licenses.go#L48
+	coverageThreshold = 75
```
kezhenxu94, line 35: And we might expose this as a configuration(with 75 as default) for advanced users?

fgksgf:
Good catch, I will handle this at this weekend. @kezhenxu94

pkg/license/identifier.go
fgksgf: I prefer to lazy init the scanner:
1. It will cost more than one second to init the scanner
2. The scanner will only be used under `dep` command

kezhenxu94: Yeah agree!

pkg/license/identifier.go
fgksgf: Ok, but this requires to change a serials of functions, I will file another PR to do this.

kezhenxu94: > Ok, but this requires to change a serials of functions, I will file another PR to do this.

Sure

kezhenxu94 (approved):

PR #114, opened 2022-06-10
Add custom license urls for identification

kezhenxu94:


assets/urls.yaml
```diff
@@ -0,0 +1,57 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
kezhenxu94, line 1: We can save some verbosity from https://github.com/google/licensecheck/pull/52 if it gets merged

fgksgf (approved):

PR #120, opened 2022-07-07
Support license expression in dep check.

jmjoy:
Fix https://github.com/apache/skywalking/issues/9319.


kezhenxu94 (approved):
@jmjoy thanks you for the use case and high quality PR, 🙇

PR #121, opened 2022-07-07
Add rust cargo support for dep command.

jmjoy:
Add rust cargo support for dep command.

### DIFF
--- a/pkg/deps/cargo.go
+++ b/pkg/deps/cargo.go
@@ -19,13 +19,13 @@ package deps
 
 import (
 	"encoding/json"
-	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"regexp"
 
 	"github.com/apache/skywalking-eyes/internal/logger"
+	"github.com/apache/skywalking-eyes/pkg/license"
 )
 
 type CargoMetadata struct {
@@ -114,10 +114,6 @@ var cargoPossibleLicenseFileName = regexp.MustCompile(`(?i)^LICENSE|LICENCE(\.tx
 // ResolvePackageLicense resolve the package license.
 // The CargoPackage.LicenseFile is generally used for non-standard licenses and is ignored now.
 func (resolver *CargoTomlResolver) ResolvePackageLicense(config *ConfigDeps, pkg *CargoPackage, report *Report) error {
-	if pkg.License == "" {
-		return fmt.Errorf("license is empty")
-	}
-
 	dir := filepath.Dir(pkg.ManifestPath)
 	logger.Log.Debugf("Directory of %+v is %+v", pkg.Name, dir)
 	files, err := os.ReadDir(dir)
@@ -127,6 +123,8 @@ func (resolver *CargoTomlResolver) ResolvePackageLicense(config *ConfigDeps, pkg
 
 	var licenseFilePath string
 	var licenseContent []byte
+
+	licenseID := pkg.License
 
 	for _, info := range files {
 		if !cargoPossibleLicenseFileName.MatchString(info.Name()) {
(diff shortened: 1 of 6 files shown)


pkg/deps/cargo.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+var cargoPossibleLicenseFileName = regexp.MustCompile(`(?i)^LICENSE|LICENCE(\.txt)?|LICENSE-.+|COPYING(\.txt)?$`)
+
+// ResolvePackageLicense resolve the package license.
+// The CargoPackage.LicenseFile is generally used for non-standard licenses and is ignored now.
+func (resolver *CargoTomlResolver) ResolvePackageLicense(config *ConfigDeps, pkg *CargoPackage, report *Report) error {
+	if pkg.License == "" {
+		return fmt.Errorf("license is empty")
+	}
```
kezhenxu94, line 119: Should be this?

```suggestion
	if pkg.License != "" {
		return pkg.License
	}
```

jmjoy: I wrote it with reference to `GoModResolver`, which returns an error, the calling place will warn and call `report.Skip`

kezhenxu94: Hi @jmjoy , I mean if `pkg.License != ""` we can just use `pkg.License` (no need to read its license file content)

jmjoy: @kezhenxu94  In fact, I don't need the --output parameter, but I saw that this parameter originally existed, so I implemented it.

kezhenxu94: OK. Then we don't need this `if` statement. This will basically report a false negative case if the `pkg.license` field is empty but there is a license file in the crate

kezhenxu94: @jmjoy I think you might want to apply this patch

```diff
diff --git a/pkg/deps/cargo.go b/pkg/deps/cargo.go
index d1057ab..e2613d2 100644
--- a/pkg/deps/cargo.go
+++ b/pkg/deps/cargo.go
@@ -19,13 +19,13 @@ package deps
 
 import (
 	"encoding/json"
-	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"regexp"
 
 	"github.com/apache/skywalking-eyes/internal/logger"
+	"github.com/apache/skywalking-eyes/pkg/license"
 )
 
 type CargoMetadata struct {
@@ -114,10 +114,6 @@ var cargoPossibleLicenseFileName = regexp.MustCompile(`(?i)^LICENSE|LICENCE(\.tx
 // ResolvePackageLicense resolve the package license.
 // The CargoPackage.LicenseFile is generally used for non-standard licenses and is ignored now.
 func (resolver *CargoTomlResolver) ResolvePackageLicense(config *ConfigDeps, pkg *CargoPackage, report *Report) error {
-	if pkg.License == "" {
-		return fmt.Errorf("license is empty")
-	}
-
 	dir := filepath.Dir(pkg.ManifestPath)
 	logger.Log.Debugf("Directory of %+v is %+v", pkg.Name, dir)
 	files, err := os.ReadDir(dir)
@@ -128,6 +124,8 @@ func (resolver *CargoTomlResolver) ResolvePackageLicense(config *ConfigDeps, pkg
 	var licenseFilePath string
 	var licenseContent []byte
 
+	licenseID := pkg.License
+
 	for _, info := range files {
 		if !cargoPossibleLicenseFileName.MatchString(info.Name()) {
 			continue
@@ -142,11 +140,17 @@ func (resolver *CargoTomlResolver) ResolvePackageLicense(config *ConfigDeps, pkg
 		break
 	}
 
+	if licenseID == "" { // If pkg.License is empty, identify the license ID from the license file content
+		if licenseID, err = license.Identify(string(licenseContent), config.Threshold); err != nil {
+			return err
+		}
+	}
+
 	report.Resolve(&Result{
 		Dependency:      pkg.Name,
 		LicenseFilePath: licenseFilePath,
 		LicenseContent:  string(licenseContent),
-		LicenseSpdxID:   pkg.License,
+		LicenseSpdxID:   licenseID,
 		Version:         pkg.Version,
 	})
 
```

jmjoy: Thanks, it's better, just push to the patch to my branch? Because I open the `Allow edits and access to secrets by maintainers`.

kezhenxu94: > Thanks, it's better, just push the patch to my branch? Because I open the `Allow edits and access to secrets by maintainers`.

Yes. Feel free to just `git apply` my patch and push in this PR. Or if you want me to edit I can do that too.

jmjoy: @kezhenxu94 Done.

PR #122, opened 2022-07-08
Add subcommand support for github actions.

jmjoy:
Usually CI not only needs to do `header check`, but also do `dep check`.

I added the `subcommand` argument to ensure compatibility, there may be a better way.


kezhenxu94 (commented):
This is not appropriate because when subcommand is `dependency`, `${{ inputs.mode }}` can be `fix`/`check`, but `dependency` has no subcommand `fix`, I was planning to have a dedicated GitHub action for dependency, in that action we can have more configurations specific to `dependency` command. If you don't mind I will open a PR to add the new GHA

jmjoy:
> This is not appropriate because when subcommand is `dependency`, `${{ inputs.mode }}` can be `fix`/`check`, but `dependency` has no subcommand `fix`, I was planning to have a dedicated GitHub action for dependency, in that action we can have more configurations specific to `dependency` command. If you don't mind I will open a PR to add the new GHA

OK.

PR #125, opened 2022-07-11
feat(header templates): add support for AGPL-3.0

elijaholmos:
adds two new header templates for the following SPDX licenses:
- [`AGPL-3.0-only`](https://spdx.org/licenses/AGPL-3.0-only.html)
- [`AGPL-3.0-or-later`](https://spdx.org/licenses/AGPL-3.0-or-later.html)

no support for [`AGPL-3.0`](https://spdx.org/licenses/AGPL-3.0.html) because it is marked as deprecated by SPDX.


kezhenxu94 (approved):
Thanks @elijaholmos

elijaholmos:
np @kezhenxu94! cheers

PR #127, opened 2022-07-21
Add MulanPSL-2.0 support.

jmjoy:


pkg/header/config.go
```diff
@@ -45,10 +45,11 @@ var (
 )
 
 type LicenseConfig struct {
-	SpdxID         string `yaml:"spdx-id"`
-	CopyrightOwner string `yaml:"copyright-owner"`
-	Content        string `yaml:"content"`
-	Pattern        string `yaml:"pattern"`
+	SpdxID            string `yaml:"spdx-id"`
+	CopyrightOwner    string `yaml:"copyright-owner"`
+	CopyrightSoftware string `yaml:"copyright-software"`
```
jmjoy, line 50: Add this field because MulanPSL-2.0 header require a [software] placeholder.

wu-sheng: What is the difference between owner and software?
These seem untypical concepts.

jmjoy: > What is the difference between owner and software? These seem untypical concepts.

[owner] is the author of the software, and [software] is the name of the software.

wu-sheng: Is it the only license having this special definition?

jmjoy: I searched for a round of license headers and found that only [gpl-3.0](https://choosealicense.com/licenses/gpl-3.0/) header receive an optional software name. 

Because the `header check `uses [FindStringIndex](https://github.com/apache/skywalking-eyes/blob/9b6b981a99fb9179f1c5aa58ac6474ea1dbdd552/pkg/header/check.go#L153) to match, and the software name of the header of the MulanPSL-2.0 protocol is just at the top, so if this placeholder is removed, the `header check` should be successful.

wu-sheng: OK, two things
1. Please name this variable as `name`. The name of the project is very clear. I think.
2. Document should be updated, https://github.com/apache/skywalking-eyes#configurations. 

@kezhenxu94 What do you think? What do you suggest to change the document index? Use `<4>` for `name`, and change the followings.

jmjoy: Or maybe move the name to the top level?

wu-sheng: @kezhenxu94 should be better to provide suggestions on codes. `CopyrightSoftware` seems not a very good name.

kezhenxu94: > OK, two things
> 
> 1. Please name this variable as `name`. The name of the project is very clear. I think.

`name` is somewhat vague to me because it's under `header.license`

```yaml
header: # <1>
  license:
    name: ... # here
```

So it may read like "header license's name". 

What about prefixing with software or project like softwareName or projectName. 


> 2. Document should be updated, https://github.com/apache/skywalking-eyes#configurations. 
> 
> 
> 
> @kezhenxu94 What do you think? What do you suggest to change the document index? Use `<4>` for `name`, and change the followings.

Both is ok. Let's use 4 for name to keep the orders

kezhenxu94: > Or maybe move the name to the top level?

Adding a top-level field is ok in this case but

- this field is only used in header command so we have no idea how to prevent it being used in other places (now or future)
- if the field is used in other places we have no idea how to differentiate them if they have different values by any chance. 
- being top-level means it has wider impact and risk of incompatibility when modified

README.md
```diff
@@ -727,7 +727,8 @@ header: # <1>
   license:
     spdx-id: Apache-2.0 # <2>
     copyright-owner: Apache Software Foundation # <3>
-    content: | # <4>
+    name: skywalking-eyes # <4>
```
wu-sheng, line 730: @kezhenxu94 mentioned about concerns about (name). Recommend (softwareName).

jmjoy: Oh, I missed it.

wu-sheng (approved):

kezhenxu94 (approved):

fgksgf (approved):

PR #130, opened 2022-08-09
Add more `.env.[mode]` support for VueJS project

rovast:
Add more `.env.[mode]` support for vuejs project. see https://cli.vuejs.org/guide/mode-and-env.html#environment-variables for more detail.


kezhenxu94 (approved):
Thanks

PR #132, opened 2022-08-31
Docker Multiple Architecture Support :fixes#9089

mohammedtabish0:
resolves apache/skywalking#9089
- Added feature to create docker multiple architectures images (linux/amd64 and linux/arm64)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,11 +76,10 @@ docker:
 
 .PHONY: docker-push
 docker-push:
-	docker buildx create --use --driver docker-container --name skywalking_eyes_main
-	docker buildx build --push --platform linux/amd64,linux/arm64 -t $(HUB)/$(PROJECT):$(VERSION) -t $(HUB)/$(PROJECT):latest .
+	docker buildx create --use --driver docker-container --name skywalking_eyes_main > /dev/null 2>&1
+	docker buildx build --push --platform linux/amd64,linux/arm64 -t $(HUB)/$(PROJECT):$(VERSION) -t $(HUB)/$(PROJECT):latest . || true
 	docker buildx rm skywalking_eyes_main
 
-.PHONY: docker-release
 docker-release: docker docker-push
 
 .PHONY: clean
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ docker:
 
 .PHONY: docker-push
 docker-push:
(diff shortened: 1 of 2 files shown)


mohammedtabish0:
@kezhenxu94 
Haven't added  `> /dev/null 2>&1` and `|| true` so that command is runnable in other platforms like windows(do we really need it?). But will cause issue if builder gets created and buildx command gets failed then running again will throw error's like builder already present.

mohammedtabish0:
@kezhenxu94 Agreed, added these commands.
Also, I have made it different from [skywalking's makefile ](https://github.com/apache/skywalking) as IMO the `|| true` should only be in the `buildx` command because if this command gets failed then cleanup of builder should run and if we are unable to create builder then we should not proceed further.

Makefile
```diff
@@ -72,15 +72,15 @@ build: windows linux darwin
...
 .PHONY: docker
 docker:
-	docker build . -t $(HUB)/$(PROJECT):$(VERSION)
+	docker build . -t $(HUB)/$(PROJECT):$(VERSION) -t $(HUB)/$(PROJECT):latest
 
 .PHONY: docker-push
 docker-push:
-	docker push $(HUB)/$(PROJECT):$(VERSION)
+	docker buildx create --use --driver docker-container --name skywalking_eyes_main
+	docker buildx build --push --platform linux/amd64,linux/arm64 -t $(HUB)/$(PROJECT):$(VERSION) -t $(HUB)/$(PROJECT):latest . || true
```
kezhenxu94, line 80: Can we save the exit code of this command if it failed and fail the whole make recipe after we clean up the builder?

This is important in the CI so we know there is something wrong instead of failed silently

mohammedtabish0: How would I save the exit code of only the first line before OR(`|| `) as this is not a pipe command and `$?` will give me the output of whole buildx command including `|| true`.

kezhenxu94: @mohammedtabish0 can you try this style?

```makefile
docker-push:
	@{ \
		echo 'the build command' ; \
		code=$$? ; \
		echo 'clean up' ; \
		exit $$code ; \
	}
```

kezhenxu94 (approved):
@mohammedtabish0 thank you for your patience and hard work!

mohammedtabish0:
@kezhenxu94 This was really fun, although I changed only three lines but got to learn a lot. Thank you so much for the patience and guidance. Hope to contribute more to the projects.

PR #133, opened 2022-09-02
feat: list files by git when possible

tisonkun:
Although we try to use ignore pattern present by a `.gitignore` file, it's possible that users have a global git ignore settings or manually ignore a file.

We can use git command to list files when runs under a git worksapce, and if there's no git workspace, we don't have to see `.gitignore`. Normally, it's via directly download and all ignored files already excluded. If users want to respect to the ignore pattern, they can simply `git init`.

cc @kezhenxu94 @wu-sheng @fgksgf @spacewander 

Reference:

* https://github.com/apache/skywalking/issues/7543
* https://github.com/apache/skywalking/issues/7538


kezhenxu94:
> Although we try to use ignore pattern present by a `.gitignore` file, it's possible that users have a global git ignore settings or manually ignore a file.
> 
> We can use git command to list files when runs under a git worksapce, 

This is a good idea, but it looks like the implementation in this PR only lists files that have been committed into the codebase, however, in most cases, I'm using the tool in the following workflow:

- In a git repo, checkout a new branch.
- Do modifications, for example, `touch a-new-file.yaml`.
- ... many other new files.
- `license-eye h f` to fix all files that are missing license headers.
- `git add . && git commit -m'...'`.

In this PR, nothing will be fixed, even if I run `license-eye h f` after `git add .`.

This also breaks users who used this tool as a git pre-commit hook.

tisonkun:
@kezhenxu94 good point. Let me try to find a way to list staged/unstaged files.

tisonkun:
@kezhenxu94 fixed at https://github.com/apache/skywalking-eyes/pull/133/commits/18af4c46d588d211d2983861689b7e7117376777.

pkg/header/check.go
```diff
@@ -63,37 +63,54 @@ func listFiles(config *ConfigHeader) ([]string, error) {
...
 			fileList = append(fileList, files...)
 		}
 	} else {
+		var candidates []string
+
+		t, _ := repo.Worktree()
+		s, _ := t.Status()
+		for file := range s {
+			candidates = append(candidates, file)
+		}
```
tisonkun, line 86: We can handle status in fine-grained. However, I just simply add all possible candidates and filter files deleted out, and duplicate them below:

```go
		seen := make(map[string]bool)
		for _, file := range candidates {
			if !seen[file] {
				seen[file] = true
				if _, err := os.Stat(file); err == nil {
					fileList = append(fileList, file)
				}
			}
		}
```

kezhenxu94 (commented):
Basically looks good to me, except that it doesn't resolve the problem as the PR description claimed, `it's possible that users have a global git ignore settings.`, the files contained in global `.gitignore` are still checked.

Reproduce:

```shell
skywalking-eyes pr/tisonkun/133 % g
On branch pr/tisonkun/133
Your branch is up to date with 'tisonkun/list-files-by-git'.

nothing to commit, working tree clean

skywalking-eyes pr/tisonkun/133 % g config --global core.excludesFile       
~/.gitignore

skywalking-eyes pr/tisonkun/133 % tail -n1 ~/.gitignore 
.tool-versions

skywalking-eyes pr/tisonkun/133 % touch .tool-versions

skywalking-eyes pr/tisonkun/133 % go run cmd/license-eye/main.go h c
INFO Totally checked 726 files, valid: 59, invalid: 1, ignored: 666, fixed: 0 
ERROR the following files don't have a valid license header: 
.tool-versions 
ERROR one or more files does not have a valid license header 
exit status 1
```

In terms of this case, it is neither better or worse than before so I'm OK if you want to stop at this point or continue to resolve that case. @tisonkun

tisonkun:
This patch fixes my case under `incubator-kvrocks`:

```
➜  incubator-kvrocks git: ~/Brittani/skywalking-eyes/bin/darwin/license-eye -c tools/ci/licenserc.yml header check
INFO Loading configuration from file: tools/ci/licenserc.yml 
INFO Totally checked 250 files, valid: 212, invalid: 0, ignored: 38, fixed: 0 
➜  incubator-kvrocks git: ~/go/bin/license-eye -c tools/ci/licenserc.yml header check
INFO Loading configuration from file: tools/ci/licenserc.yml 
INFO Totally checked 4279 files, valid: 212, invalid: 2045, ignored: 2022, fixed: 0 
INFO GITHUB_TOKEN is not set, license-eye won't comment on the pull request 
ERROR the following files don't have a valid license header: 
cmake-build-debug/_deps/gtest-src/googletest/scripts/test/Makefile
cmake-build-debug/_deps/luajit-src/src/Makefile
...
cmake-build-debug/_deps/lz4-src/contrib/snap/snapcraft.yaml 
ERROR one or more files does not have a valid license header
```

... where:

```
cat $(git config --global core.excludesFile)
# JetBrains
.idea
*.iml
cmake-build-*

# VS Code
.vscode

# macOS
.DS_Store

# Ruby
.ruby-version
```

I'll try to dig it out a bit :)

tisonkun:
@kezhenxu94 I think we can handle the returned `FileStatus`. Previously, I assume that all ignored files don't occur in the status list, but it seems not true :)

tisonkun:
@kezhenxu94 it's a `go-git` wart. Workaround at 2687d4ee5f20e4589a79367536196485176be3be.

There're still some issues about `.gitignore` files in nested path https://github.com/go-git/go-git/pull/522. But previously we also only handle the gitignore file at root path of the repo. If someone else asks for it, they can nudge the upstream fix or switch to a patching mod replacement. I'll leave it as is for now.

tisonkun:
> This patch fixes my case

I spot that it's about another issue on the master branch. I'll create an issue for it.

pkg/header/check.go
```diff
@@ -28,92 +30,144 @@ import (
...
+			return nil, err
+		}
+		if err := tree.Files().ForEach(func(file *object.File) error {
+			if file == nil {
+				return errors.New("file pointer is nil")
+			}
+			if _, err := os.Stat(file.Name); err == nil {
+				candidates = append(candidates, file.Name)
+			}
+			return nil
```
kezhenxu94, line 112: I prefer to keep the previous logic so when `os.Stat` reports an error we won't leave out a file, (we don't care the error, tho). 

```suggestion
			if _, err := os.Stat(file.Name); err != nil {
				return err
			}
			candidates = append(candidates, file.Name)
			return nil
```

pkg/header/check.go
```diff
@@ -28,92 +30,144 @@ import (
...
+		}
 
-func glob(pattern string) (matches []string, err error) {
-	if pattern == "." {
-		return doublestar.Glob("./")
+		seen := make(map[string]bool)
+		for _, file := range candidates {
+			if !seen[file] {
+				seen[file] = true
+				if _, err := os.Stat(file); err == nil {
```
kezhenxu94, line 121: Same here

kezhenxu94 (approved):
There is still a culprit in the go-git upstream, it doesn't expand the tilde (`~`) for the `core.excludesfile`, which I think it should, so this still doesn't solve my problem.

```shell
skywalking-eyes pr/tisonkun/133 % g config --global core.excludesfile
~/.gitignore
```

Also left some minor comment, the PR is good enough for me.

pkg/header/check.go
tisonkun: It's possible that a file is removed in staging but not committed. Thus, files listed by `git ls-files` or `git status` can be deleted. It's not an error case.

It's possible we skip return error if `os.IsNotExist(err) == true`.

WDYT?

pkg/header/check.go
tisonkun: See https://github.com/apache/skywalking-eyes/pull/133/files#r962437870. Although, we can merge these two blocks.

pkg/header/check.go
kezhenxu94: > It's possible that a file is removed in staging but not committed. Thus, files listed by `git ls-files` or `git status` can be deleted. It's not an error case.
> 
> It's possible we skip return error if `os.IsNotExist(err) == true`.
> 
> WDYT?

OK I agree, let's add the if condition.

tisonkun: Updated at https://github.com/apache/skywalking-eyes/pull/133/commits/8565a1f85bc34debafd764a39cc1c5bb287ba0b7.

tisonkun:
> it doesn't expand the tilde (`~`) for the `core.excludesfile`

I dug the code a bit. Finally, I notice that it's the Golang stdlib that causes the issue:

```go
func main() {
	_, err := os.OpenFile("~/.gitignore", 0, 0666)
	fmt.Printf("%v", err)
}

// open ~/.gitignore: no such file or directory
```

kezhenxu94 (approved):

tisonkun:
Setup a cross-reference to https://github.com/go-git/go-git/issues/578. If upstream resolves this issue, we can upgrade to fix the `~` in path issue.

PR #134, opened 2022-09-03
Polish maven test for convenient debug

kezhenxu94:
Now we can debug the identification in the test, which makes it simple to debug why we failed to identify the license of a dependency

### DIFF
--- a/pkg/deps/maven_test.go
+++ b/pkg/deps/maven_test.go
@@ -19,13 +19,14 @@ package deps_test
 
 import (
 	"bufio"
+	"embed"
 	"fmt"
+	"io/fs"
 	"os"
-	"os/exec"
 	"path/filepath"
+	"strings"
 	"testing"
 
-	"github.com/apache/skywalking-eyes/internal/logger"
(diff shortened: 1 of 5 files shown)


wu-sheng:
I can't follow the PR description. What does this change mean?

pkg/deps/maven_test.go
```diff
@@ -106,14 +132,22 @@ func TestResolveMaven(t *testing.T) {
 				<artifactId>skywalking-sharing-server-plugin</artifactId>
 				<version>8.6.0</version>
 			</dependency>
+			<dependency>
+				<groupId>com.fasterxml.jackson.datatype</groupId>
+				<artifactId>jackson-datatype-jsr310</artifactId>
+				<version>2.13.3</version>
+			</dependency>
```
kezhenxu94, line 139: > I can't follow the PR description. What does this change mean?

Previously when the tool failed to identify a dependency's license, we have to create a demo maven project and run the tool in debug mode, now we can just put the dependency in the test codes and debug the test, like this.

wu-sheng (approved):

PR #135, opened 2022-09-06
Switch to npm ci for reliable builds

kezhenxu94:
According to https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable, one should prefer `npm ci` over `npm install` for more reliable builds, so all the versions can be locked exactly unless we explicitly upgrade them.

### DIFF
--- a/pkg/deps/npm.go
+++ b/pkg/deps/npm.go
@@ -127,7 +127,9 @@ func (resolver *NpmResolver) NeedSkipInstallPkgs() bool {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


wu-sheng (dismissed):

wu-sheng:
Should we use `npm ci` in the main repo to package?

kezhenxu94:
> Should we use `npm ci` in the main repo to package?

We need to fix the UI repo first https://github.com/apache/skywalking-booster-ui/pull/154

wu-sheng (approved):

PR #136, opened 2022-09-07
Fix optional dependencies are not excluded

kezhenxu94:
Optional dependencies should be excluded but they are not because of regexp missing the optional part

```
[INFO] +- org.mariadb.jdbc:mariadb-java-client:jar:2.7.2:provided (optional) 
```


fgksgf (approved):

slievrly:
@kezhenxu94 get the latest code, `make build`, and the test case passes.

PR #137, opened 2022-09-08
Fix exclude not work for transitive dependencies and add recursive config

kezhenxu94:
`dependency.excludes` doesn't work for transitive dependencies so this might bring some dependencies that users want to exclude.

This patch also adds a `recursive` config so users can decide whether to only exclude one dependency or also exclude all dependencies brought by this dependency.

/cc @jmjoy currently it seems Rust (cargo.go) only resolve direct dependencies, not including transitive dependencies, so we might miss some transitive dependencies in the license report.

/cc and thanks to @slievrly for reporting this bug

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,13 +31,13 @@ header:
 
   comment: on-failure
 
-# If you want to check dependencies' license compatibility, uncomment the following section
-# dependency:
-#   files:
-#     - pom.xml           # If this is a maven project.
-#     - Cargo.toml        # If this is a rust project.
-#     - package.json      # If this is a npm project.
-#     - go.mod            # If this is a Go project.
+# If you don't want to check dependencies' license compatibility, remove the following part
+dependency:
+  files:
+    - pom.xml           # If this is a maven project.
+    - Cargo.toml        # If this is a rust project.
+    - package.json      # If this is a npm project.
+    - go.mod            # If this is a Go project.
 ```
 
 #### Check License Headers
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -31,13 +31,13 @@ header:
...
 
-# If you want to check dependencies' license compatibility, uncomment the following section
-# dependency:
-#   files:
-#     - pom.xml           # If this is a maven project.
-#     - Cargo.toml        # If this is a rust project.
-#     - package.json      # If this is a npm project.
-#     - go.mod            # If this is a Go project.
+# If you don't want to check dependencies' license compatibility, remove the following part
+dependency:
```
kezhenxu94, line 35: Previously when users uncomment the the codes they might make the indentation wrong, so I uncomment these by default and users can opt-out by deleting them

README.md
```diff
@@ -202,193 +202,61 @@ license-eye -c test/testdata/.licenserc_for_test_check.yaml dep resolve -o ./dep
...
-
-ERROR failed to identify the licenses of following packages (7):
-github.com/gogo/protobuf
-github.com/kr/logfmt
-github.com/magiconair/properties
-github.com/miekg/dns
-github.com/pascaldekloe/goe
-github.com/russross/blackfriday/v2
-gopkg.in/check.v1
+INFO Loading configuration from file: .licenserc.yaml
```
kezhenxu94, line 205: Now that we improve the correctness, I update the example to reflect that.

wu-sheng (approved):

slievrly:
@kezhenxu94  test has passed.

PR #141, opened 2022-09-26
fix: not ignore *.txt to make sure files like CMakeLists.txt can be checked

acelyc111:


kezhenxu94 (approved):
This looks to be a minor breaking changes, but users can easily add txt to their ignore path so this LGTM

PR #46, opened 2021-07-22
Enhance Go dependency resolver to resolve built-in and non-standard Go modules

kezhenxu94:
This patch enhances the Go dependency resolver, previously for some dependencies we are not able to resolve them, now it's able to find all modules.

example output

```text
go run cmd/license-eye/main.go d r 
INFO GITHUB_TOKEN is not set, license-eye won't comment on the pull request 
INFO Loading configuration from file: .licenserc.yaml 
WARNING Failed to resolve the license of dependency: github.com/pascaldekloe/goe cannot identify license content 
Dependency                                         |        License
-------------------------------------------------- | --------------
github.com/apache/skywalking-eyes/license-eye      |     Apache-2.0
cloud.google.com/go                                |     Apache-2.0
cloud.google.com/go/bigquery                       |     Apache-2.0
cloud.google.com/go/datastore                      |     Apache-2.0
cloud.google.com/go/firestore                      |     Apache-2.0
cloud.google.com/go/pubsub                         |     Apache-2.0
cloud.google.com/go/storage                        |     Apache-2.0
dmitri.shuralyov.com/gpu/mtl                       |   BSD-3-Clause
github.com/BurntSushi/xgb                          |   BSD-3-Clause
github.com/OneOfOne/xxhash                         |     Apache-2.0
github.com/alecthomas/template                     |   BSD-3-Clause
github.com/armon/circbuf                           |            MIT
github.com/armon/go-metrics                        |            MIT
github.com/armon/go-radix                          |            MIT
github.com/beorn7/perks                            |            MIT
github.com/bgentry/speakeasy                       |            MIT
github.com/bketelsen/crypt                         |            MIT
github.com/bmatcuk/doublestar/v2                   |            MIT
github.com/cespare/xxhash                          |            MIT
github.com/client9/misspell                        |            MIT
github.com/coreos/bbolt                            |            MIT
github.com/coreos/etcd                             |     Apache-2.0
github.com/coreos/go-semver                        |     Apache-2.0
github.com/coreos/go-systemd                       |     Apache-2.0
github.com/coreos/pkg                              |     Apache-2.0
github.com/cpuguy83/go-md2man/v2                   |            MIT
github.com/davecgh/go-spew                         |            ISC
github.com/dgrijalva/jwt-go                        |            MIT
github.com/dgryski/go-sip13                        |            MIT
github.com/fatih/color                             |            MIT
github.com/fsnotify/fsnotify                       |   BSD-3-Clause
[...]

PR #142, opened 2022-10-27
fix license header normalizer

xiaoyawei:
## Why do we need this fix?

As we are using `license-eyes` to check license header in our open-sourced projects, we notice that it can help detect 
unexpected changes in the license header, *except the copyright line*. In my opinion, it's also necessary to protect the copyright line from unintentional changes.

## How was this issue introduced?

https://github.com/apache/skywalking-eyes/pull/46 introduced this line processor to ignore copyright line, mostly for the purpose of dependency resolution. However, in https://github.com/apache/skywalking-eyes/pull/107 starts using Google's license check for this purpose, and this line processor only works to ignore copyright line when checking / fixing license header.

## Reviewer

@kezhenxu94

### DIFF
--- a/pkg/license/norm.go
+++ b/pkg/license/norm.go
@@ -233,12 +233,6 @@ var (
 			regexp.MustCompile(`(?m)\s+[*]$`),
 			" ",
 		},
-		// Copyright (c) .....
-		// © Copyright .....
-		{
-			regexp.MustCompile(`(?m)^\s*([cC©])?\s*Copyright (\([cC©]\))?.+$`),
-			"",
-		},
 		// Portions Copyright (C) ...
 		{
 			regexp.MustCompile(`(?m)^\s*Portions Copyright (\([cC©]\))?.+$`),


kezhenxu94 (approved):

tisonkun:
This commit causes a previously passed CI to fail. You may re-check the diff: https://github.com/tisonkun/failpoints/actions/runs/3342840329/jobs/5535474781

xiaoyawei:
@tisonkun your CI failures look reasonable to me: its copyright line in license header does not match the specs in https://github.com/tisonkun/failpoints/blob/main/tools/ci/licenserc.yml

kezhenxu94:
I knew this was a breaking change, but it breaks something that was wrongly configured, and the previous successful CI should have been failed.

In your case, using spdx-id whose content has variables should be substituted, so you might need to provide `copyright-owner`, I know that also failed because of a bug https://github.com/apache/skywalking-eyes/pull/143, the variables are not substituted previously.

kezhenxu94:
@tisonkun with #143 and the following patch to your repo, CI should pass

```diff
diff --git a/tools/ci/licenserc.yml b/tools/ci/licenserc.yml
index 7122e11..2c56a5d 100644
--- a/tools/ci/licenserc.yml
+++ b/tools/ci/licenserc.yml
@@ -15,6 +15,7 @@
 header:
   license:
     spdx-id: Apache-2.0
+    copyright-owner: tison <wander4096@gmail.com>
   paths:
     - '**/*.java'
     - '**/*.yml'
```

Let me know if that fix makes sense to you.

tisonkun:
@kezhenxu94 fair enough. Apply and fix the issue :)

PR #151, opened 2023-02-24
Fix GitHub Actions wrong path

kezhenxu94:


wu-sheng (dismissed):

kezhenxu94:
> I have some issue with `skywalking-eyes`, does it relate with this PR, follow is error message:
> 
> ```
> Error: Process completed with exit code 2.
> ```

Maybe, can you please share which repo you're using? Also we recommend not to use `@main` tag directly.

wu-sheng (approved):

PR #163, opened 2023-05-16
dependency resolve with default template and specified output of license

crholm:
The cli semantic for flags is somewhat cumbersome, unexpected behavior and does not lend it self well for github actions when we want to automatically generate a `LICENSE` file.  

`license-eye dependency resolve` has the `--summary` flag which takes a path to a go template file, and emits a LICENCE file in the template dir and only if specified. 

The `--summary` flag thus has three functions, providing a template for rendering, specifying path to output and toggling functionality for outputting a LICENSE file or not.

My suggestion, to keep backwards compatibility, is to add a new flag, `--license`, which specifies where to write the license file and adding a default template. `--summary` will retains its behavior for projects depending on it, while allowing a default template and a specific path to be used.

The PR does the following
* adds the flag --license, to the command `dependency resolve`, which takes a path to  where the LICENSE content shall be written
* adds a default template for LICENSE which is used if `--license` is specified but not `--summary`
* adds the `inputs.flags` to github action `dependency/action.yml` in order to be able to specify flags to generate a LICENSE file through github actions
*  updating README with information of the new flag and github action


kezhenxu94 (commented):
Looks good to me, @crholm would you fix the linting errors so the CI can pass?

crholm:
Linting fixed, 

An extra linting error had to be fixed, probably new check in linter
https://github.com/apache/skywalking-eyes/pull/163/files#diff-8a27b8884c79bafd31eba6126f1baf57c0beecc820b372f2aa5fd74b16d1ec05R77

Header exclusion also was added to have `make all` to work properly 
https://github.com/apache/skywalking-eyes/pull/163/files#diff-f8a05717594bf148241760c18447442978fdc9db03a4aae93739529e0c124f8aR74

kezhenxu94 (approved):
Thank you @crholm

PR #166, opened 2023-08-16
Remove "portions copyright" header normalizer

antgamdia:
Following up on the changes performed at https://github.com/apache/skywalking-eyes/pull/142, this PR is removing another regex that is causing the whole copyright line to be removed by the normalizer (due to the `.+$` match) if the line contains `Portions Copyright`.

 closes https://github.com/apache/skywalking/issues/11218

cc @kezhenxu94

### DIFF
--- a/pkg/license/norm.go
+++ b/pkg/license/norm.go
@@ -233,11 +233,6 @@ var (
 			regexp.MustCompile(`(?m)\s+[*]$`),
 			" ",
 		},
-		// Portions Copyright (C) ...
-		{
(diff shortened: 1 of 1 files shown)


kezhenxu94:
I'll cancel the 0.5.0 and re-initiate the release with this patch in

kezhenxu94 (approved):

wu-sheng:
> I'll cancel the 0.5.0 and re-initiate the release with this patch in

@kezhenxu94 I noticed 0.5.0 tag still exists for cancelled release. Would you replace it with the latest commit and initial a new release?

PR #171, opened 2023-10-31
update: add weak-compatible to dependency check

Two-Hearts:
This PR is based on this discussion: https://github.com/apache/skywalking/discussions/11391.
A new list called `weak-compatible` is added to the [skywalking-eyes compatibility list](https://github.com/apache/skywalking-eyes/blob/main/assets/compatibility/Apache-2.0.yaml). 
A new flag called `weak-compatible` is added to the `dependency check` command.

### DIFF
--- a/commands/deps_check.go
+++ b/commands/deps_check.go
@@ -29,7 +29,7 @@ import (
 var weakCompatible bool
 
 func init() {
-	DepsCheckCommand.PersistentFlags().BoolVarP(&weakCompatible, "weak-compatible", "w", false, "if set, include the weak-compatible list in dependencies checking")
+	DepsCheckCommand.PersistentFlags().BoolVarP(&weakCompatible, "weak-compatible", "w", false, "if set to true, treat the weak-compatible licenses as compatible in dependencies check. Note: when set to true, make sure to manually confirm that weak-compatible licenses are used under the required conditions.")
 }
 
 var DepsCheckCommand = &cobra.Command{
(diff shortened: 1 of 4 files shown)


wu-sheng:
Thanks for raising this. Please fix the CI first.

Two-Hearts:
Working on the CI, looks like a lint problem.

commands/deps_check.go
```diff
@@ -26,6 +26,12 @@ import (
 	"github.com/apache/skywalking-eyes/pkg/deps"
 )
 
+var weakCompatible bool
+
+func init() {
+	DepsCheckCommand.PersistentFlags().BoolVarP(&weakCompatible, "weak-compatible", "w", false, "if set, include the weak-compatible list in dependencies checking")
```
kezhenxu94, line 32: To me, "include the weak-compatible list in dependencies checking" doesn't indicate the "checking" accepts or rejects the list in dependencies, which might be ambiguous for users, what do you think if we make it more explicit in the description, and add a note, for example:

if set to true, treat the `weak-compatible` licenses as compatible in dependencies check. Note: when set this to true, make sure to manually confirm that you are using the `weak-compatible` licenses under the required conditions.

kezhenxu94 (commented):
A tiny suggestion about the description of the new command option, others look good to me, LGTM after CI is passed

commands/deps_check.go
Two-Hearts: Thanks @kezhenxu94, I updated the description based on your example. CI fixed.

wu-sheng (approved):
Good to have this. Wait for @kezhenxu94 on codes.

kezhenxu94 (approved):
Thank you! @Two-Hearts

PR #183, opened 2024-05-14
Update Apache-2.0.yaml to add BSD-2-Clause-Views

spacewander:
According to the https://spdx.org/licenses/BSD-2-Clause-Views.html, the license is 'identical to BSD-2-Clause with the addition of the "views and conclusions" sentence at the end'

### DIFF
--- a/assets/compatibility/Apache-2.0.yaml
+++ b/assets/compatibility/Apache-2.0.yaml
@@ -22,6 +22,7 @@ compatible:
   - PHP-3.01
   - BSD-3-Clause
   - BSD-2-Clause
+  - BSD-2-Clause-Views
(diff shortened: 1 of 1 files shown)


wu-sheng:
I think whether this is compatible with APLv2, it should be determined by ASF legal team.

spacewander:
Is there someone from ASF legal team we can consult?

wu-sheng:
We are following here, https://www.apache.org/legal/resolved.html

> Raise the topic on legal-discuss@ or via a JIRA issue if you have doubt over whether a work falls in the public domain.

This is recommended from that page.

wu-sheng (approved):

PR #207, opened 2025-09-13
Fixed nil dependency panic in Ruby Gemfile.lock resolver

pboling:
- now treats missing specs as unresolved licenses
- adds them to the invalid list
- License fetch logic handles empty gem versions gracefully
- tests updated
  - handle missing specs gracefully with mocked HTTP responses
  - imports and packages were refactored to enable mocking and improve test structure


pboling:
@wu-sheng @kezhenxu94 I think this may be sufficient.  I don't see any other changes needed.

kezhenxu94 (approved):
Looks good to me, thanks!

PR #208, opened 2025-09-13
GemfileLockResolver adjusted to exclude all specs when runtime dependencies are empty

pboling:
- regression test: Ruby library projects with no runtime dependencies do not include development dependencies

### DIFF
--- a/pkg/deps/ruby.go
+++ b/pkg/deps/ruby.go
@@ -70,12 +70,11 @@ func (r *GemfileLockResolver) Resolve(lockfile string, config *ConfigDeps, repor
 		if err != nil {
 			return err
 		}
-		if len(runtimeRoots) == 0 {
-			// Fallback: if not found, use DEPENDENCIES from lockfile
-			roots = deps
-		} else {
-			roots = runtimeRoots
-		}
+		// Use only runtime dependencies from gemspec(s);
+		// Do not fallback to DEPENDENCIES
+		// A gem library inherits runtime dependencies ONLY from the gemspec.
+		// All other dependencies are development only.
+		roots = runtimeRoots
 	} else {
 		// App: all declared dependencies are relevant
 		roots = deps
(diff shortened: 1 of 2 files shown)


pboling:
@kezhenxu94 @wu-sheng this is done!

pboling:
@kezhenxu94 This is so cool!  I am not sure what I need to do next though. It is still failing.
```
ERROR the following licenses are unknown or incompatible with the main license, please check manually: MIT
Dependency       | License
---------------- | -------
faraday          |     MIT
multi_xml        |     MIT
version_gem      |     MIT
faraday-net_http |     MIT
jwt              |     MIT
rack             |     MIT
snaky_hash       |     MIT
hashie           |     MIT
git              |     MIT
json             |    Ruby 
ERROR one or more errors occurred checking license compatibility 
Error: Process completed with exit code 1.
```

Ref: https://github.com/ruby-oauth/oauth2/actions/runs/17692969800/job/50291221524#step:3:129

IMO, these licenses are all compatible with MIT, especially the ones that *are* MIT. :laughing: 

Perhaps we need to configure the MIT license as compatible with other licenses?

pboling:
Ruby license is a license, and many of the standard ruby libraries (default, bundled, core) use it!
https://stdgems.org/compare/
I'll work on another PR :)
Did you see my request to join the slack?

pboling:
Also, the ruby license doesn't have a standard license header... is that a problem?

kezhenxu94:
> Also, the ruby license doesn't have a standard license header... is that a problem?

In terms of dependency license check, that’s not a problem. For license header, we can just use the spdx id as the license header.

pboling:
> [apache.org/legal/resolved.html#category-b](https://www.apache.org/legal/resolved.html#category-b) this is ASF specific, the weak-compatibility might not be applicable to other license, though.

1. True, but it is the best information we have to build on, and since this project isn't a lawyer it must sit on someone's shoulders.
2. Logically if an Apache project can include code from any and all licenses from category A in a compatible manner, it means they are all compatible with each other according to their definition of compatible.
3. If we don't have a better definition of compatible we should use theirs.
4. Similar with the weak-compatible.  If any and all of the weak compatible licenses can be included (for binary code) in an Apache project, it means they must be, according to their definition, compatible with each other, not just with Apache.

> I don’t see any email, did you sent to [dev@skywalking.apache.org](mailto:dev@skywalking.apache.org)?

Yes. I will send it again!

pboling:
Sent another request. Subject `Request to join SkyWalking slack`, and it is from my public gmail: peter.boling `at` gmail `dot` com

PR #266, opened 2026-03-16
Add support for Ruby, ERB and slim

afriqs:
See https://github.com/apache/skywalking/issues/7744 :sweat_smile:


kezhenxu94:
Please fix the CI

kezhenxu94:
@copilot fix the CI

assets/styles.yaml
```diff
@@ -115,3 +120,8 @@
   start: '"""'
   middle: ~
   end: '"""'
+
+- id: Slash
+  start: '^\/'
+  middle: '^\/'
+  end: '^\/'
```
wu-sheng:
```suggestion
  start: '/'
  middle: '/'
  end: '/'
```
The comment style markers are literal texts joined by `GenerateLicenseHeader`, not regular expressions, so these should be the literal `/`.

wu-sheng (commented):
I checked out this PR locally to debug the CI failure. There are two separate problems, and fixing only the first one is not enough:

**1. The `Slash` comment style must be literal text, not a regular expression** (the inline suggestion below). The style markers are joined literally by `GenerateLicenseHeader`, which is why the generated headers currently contain a literal `^\/`. That's what breaks `TestFix/test.slim` and `TestRewriteContent/Slim*`.

**2. `header check` cannot recognize single-`/` comments at all**, which is why `TestCheckFile/WithLicense/.../testcase.slim` fails even after reverting to a literal `/` (this is probably what prompted the `^\/` attempt in 23c1445). The license normalizer in `pkg/license/norm.go` strips `//`, `/*`, `*/` etc. from headers before matching, but has no rule for a single `/`, so Slim headers never match the configured license. Please add this rule (it must come after the `/*` family so it doesn't interfere):

```diff
--- a/pkg/license/norm.go
+++ b/pkg/license/norm.go
@@ -52,6 +52,7 @@
 		regexp.MustCompile(`(?m)^\s*/\*+`), // /*
 		regexp.MustCompile(`(?m)^\s*\*+/`), //  */
 		regexp.MustCompile(`(?m)^\s*\*+`),  //  *
+		regexp.MustCompile(`(?m)^\s*/`),    //  / e.g. Slim, must come after the /* and // patterns
 
 		regexp.MustCompile(`(?m)^\s*<!--+`), // <!--
 		regexp.MustCompile(`(?m)^\s*--+>`),  // -->
```

With both changes applied I verified locally on a clean checkout of this branch (and on its merge with current `main`): the full test suite passes, `golangci-lint` is clean, and an end-to-end round-trip works — `header fix` on fresh `.slim`/`.erb`/`.rb` files writes correct headers and `header check` then validates all of them.

Since "Allow edits by maintainers" is disabled on the fork, I can't push these two changes to this branch myself — please apply the suggestion below and the diff above, and the CI should turn green.

assets/styles.yaml
afriqs: Thanks for troubleshooting it. Just pushed the updated commits.

kezhenxu94 (approved):

PR #279, opened 2026-07-28
fix: don't treat mid-file #! strings as Python shebangs

cmaluend:
## Summary
- The Python comment-style `after` pattern used `^*#!`, which could match `#!` mid-line (e.g. inside a string), so `header fix` inserted the license in the wrong place.
- Anchor shebang matches to the start of a line (`^#!`) in `PythonStyle` / `PythonDocStringStyle`.
- Add rewrite unit tests for a mid-file shebang-like string, with and without a real first-line shebang.

### Example
Before this fix, `header fix` could insert the license after a shebang-like string in the middle of the file:

```python
def some_function():
    print(
        """#!/usr/bin/env python3
        print("Hello, World!")
        """)
```

became something like:

```python
def some_function():
    print(
        """#!/usr/bin/env python3
# Licensed to the Apache Software Foundation (ASF) under one
# ...
        print("Hello, World!")
        """)
```

Now the license is inserted at the top of the file (or after a real first-line shebang, when present).

### DIFF
--- a/assets/styles.yaml
+++ b/assets/styles.yaml
@@ -116,7 +116,7 @@
 
 - id: PythonDocStringStyle
   # (interpreter binary and encoding comment) | (only interpreter binary) | (only encoding comment)
-  after: '(?m)(^#!.+$\n^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)|(^#!.+$)|(^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)'
+  after: '(?m)(\A#!.+$\n^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)|(\A#!.+$)|(\A[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)'
   start: '"""'
   middle: ~
   end: '"""'
--- a/assets/styles.yaml
+++ b/assets/styles.yaml
@@ -21,7 +21,7 @@
   end: '//'
 
 - id: Hashtag
-  after: '(?m)^*#!.+$'
+  after: '(?m)\A#!.+$'
   start: '#'
   middle: '#'
   end: '#'
--- a/assets/styles.yaml
+++ b/assets/styles.yaml
@@ -109,7 +109,7 @@
 
 - id: PythonStyle
   # (interpreter binary and encoding comment) | (only interpreter binary) | (only encoding comment)
-  after: '(?m)(^#!.+$\n^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)|(^#!.+$)|(^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)'
+  after: '(?m)(\A#!.+$\n^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)|(\A#!.+$)|(\A[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)'
   start: '#'
   middle: '#'
   end: '#'
--- a/assets/styles.yaml
+++ b/assets/styles.yaml
@@ -109,14 +109,14 @@
 
 - id: PythonStyle
   # (interpreter binary and encoding comment) | (only interpreter binary) | (only encoding comment)
-  after: '(?m)(^*#!.+$\n^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)|(^*#!.+$)|(^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)'
+  after: '(?m)(^#!.+$\n^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)|(^#!.+$)|(^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)'
   start: '#'
   middle: '#'
   end: '#'
 
 - id: PythonDocStringStyle
   # (interpreter binary and encoding comment) | (only interpreter binary) | (only encoding comment)
-  after: '(?m)(^*#!.+$\n^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)|(^*#!.+$)|(^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)'
+  after: '(?m)(^#!.+$\n^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)|(^#!.+$)|(^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)'
   start: '"""'
   middle: ~
   end: '"""'
--- a/pkg/header/fix_test.go
+++ b/pkg/header/fix_test.go
@@ -215,6 +215,46 @@ EOF
 }
 `},
 		{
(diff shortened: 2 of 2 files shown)


assets/styles.yaml
```diff
@@ -109,14 +109,14 @@
 
 - id: PythonStyle
   # (interpreter binary and encoding comment) | (only interpreter binary) | (only encoding comment)
-  after: '(?m)(^*#!.+$\n^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)|(^*#!.+$)|(^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)'
+  after: '(?m)(^#!.+$\n^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)|(^#!.+$)|(^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)'
```
wu-sheng: **PythonStyle** — `^#!` with `(?m)` still matches the start of **any** line, and `rewriteContent` uses `FindIndex`, which takes the leftmost match anywhere in the file. So this narrows the bug rather than fixing it — a `#!` at column 0 inside a mid-file string or comment is still treated as a shebang:

```python
x = """
#!/usr/bin/env python3
# Licensed to the Apache Software Foundation (ASF) under one   <-- still injected mid-string
hello
"""
```

A shebang is only meaningful on line 1, so the anchor should be `\A`. Same for the encoding branch — PEP 263 only honours a `coding:` declaration on line 1 (line 2 is already covered by the shebang branch above it), but today `^[ \t\f]*#.*?coding[:=]` matches an indented `# coding:` comment at any depth anywhere in the file.

Keep `(?m)` — it is still needed for the `$` in each branch.

```suggestion
  after: '(?m)(\A#!.+$\n^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)|(\A#!.+$)|(\A[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)'
```

Verified locally: `go test ./...` passes with this, including the existing `Python with encoding`, `Python with interpreter binary`, and `Python with interpreter binary and encoding` cases.

wu-sheng (commented):
Nice catch on `^*`. Worth spelling out for future readers *why* it was wrong: in RE2 `^*` is zero-or-more of the `^` assertion, which makes the anchor **optional**, so `#!` matches anywhere on a line. The direction here is right and the added tests are genuine regressions.

Two things before this is complete.

### 1. `Hashtag` has the identical typo and isn't fixed here

`assets/styles.yaml:24` (outside the diff, so I can't leave an inline suggestion on it):

```yaml
- id: Hashtag
  after: '(?m)^*#!.+$'
```

That style backs 23 languages — including **Shell, Perl, Ruby, Tcl, R, Makefile, Dockerfile, YAML** — i.e. exactly the family where shebangs matter most. On a shell heredoc:

```
Hashtag `^*#!.+$`   -> match="#!/bin/bash is just text here" at [24 53]
        `\A#!.+$`   -> no match
```

Please apply the same fix in this PR:

```yaml
- id: Hashtag
  after: '(?m)\A#!.+$'
```

`(?m)` has to stay — without it `$` only matches end-of-text, so `.+$` would require the file to consist of nothing but the shebang line.

### 2. `^#!` narrows the bug rather than fixing it

Details inline on `assets/styles.yaml`. Short version: `(?m)^` matches the start of every line and `FindIndex` is leftmost-anywhere, so a column-0 `#!` mid-file is still mistaken for a shebang. Behaviour I measured against a replica of `rewriteContent`:

| Input shape | old `^*#!` | this PR `^#!` | `\A#!` |
|---|---|---|---|
| `#!` indented inside a mid-file string | matches | no match | no match |
| `#!` at column 0 inside a mid-file string | matches | **matches** | no match |
| `#!keep this` column-0 comment mid-file | matches | **matches** | no match |
| real first-line shebang | matches | matches | matches |
| shebang + PEP 263 coding line | matches | matches | matches |
| indented `# -*- coding: utf-8 -*-` mid-file | matches | **matches** | no match |

Given that, the PR description's claim — "the license is inserted at the top of the file (or after a real first-line shebang, when present)" — is stronger than what the change currently guarantees. Either tighten to `\A` (suggested inline; verified green on `go test ./...`) or soften the description.

assets/styles.yaml
```diff
@@ -109,14 +109,14 @@
...
   start: '#'
   middle: '#'
   end: '#'
 
 - id: PythonDocStringStyle
   # (interpreter binary and encoding comment) | (only interpreter binary) | (only encoding comment)
-  after: '(?m)(^*#!.+$\n^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)|(^*#!.+$)|(^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)'
+  after: '(?m)(^#!.+$\n^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)|(^#!.+$)|(^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)'
```
wu-sheng: **PythonDocStringStyle** — same fix as above; these two `after` patterns need to stay in sync.

```suggestion
  after: '(?m)(\A#!.+$\n^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)|(\A#!.+$)|(\A[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*$)'
```

pkg/header/fix_test.go
```diff
@@ -133,6 +133,48 @@ if __name__ == '__main__':
...
+# Apache License 2.0
+#   http://www.apache.org/licenses/LICENSE-2.0
+# Apache License 2.0
+
+def some_function():
+    print(
+        """#!/usr/bin/env python3
+        print("Hello, World!")
+        """)
 `},
```
wu-sheng, line 256: Thanks for adding real regression tests — I confirmed they genuinely fail on `main` (reverting only `assets/styles.yaml` to 6e368ec reproduces the corrupt output from the PR description).

Both new cases only cover the *indented* variant. Could you also cover the column-0 variant and a `Hashtag`-style language? Both of these fail on this branch as-is, and pass once the `\A` anchors are applied:

```suggestion
`},
		{
			name:  "Python with column-0 shebang-like string mid file",
			style: comments.FileCommentStyle("test.py"),
			content: `x = """
#!/usr/bin/env python3
hello
"""
`,
			licenseHeader: getLicenseHeader("test.py", t.Error),
			expectedContent: `# Apache License 2.0
#   http://www.apache.org/licenses/LICENSE-2.0
# Apache License 2.0

x = """
#!/usr/bin/env python3
hello
"""
`},
		{
			name:  "Shell with column-0 shebang-like string mid file",
			style: comments.FileCommentStyle("test.sh"),
			content: `foo() {
  cat <<EOF
#!/bin/bash is just text
EOF
}
`,
			licenseHeader: getLicenseHeader("test.sh", t.Error),
			expectedContent: `# Apache License 2.0
#   http://www.apache.org/licenses/LICENSE-2.0
# Apache License 2.0

foo() {
  cat <<EOF
#!/bin/bash is just text
EOF
}
`},
```

wu-sheng:
Hi, could you verify the new proposal?

cmaluend:
@wu-sheng Thanks for your feedback.
* I merged your suggestion
* I fixed the HashTag style
* I added more unit tests to cover new cases

wu-sheng (approved):

cmaluend:
@wu-sheng 
Do you know when we will have a new release with this change?

wu-sheng:
I will fix the docker push, and you could use the {commit_id} version by then.

wu-sheng:
https://github.com/apache/skywalking-eyes/pull/279 fixed the image pushing. The latest could be found here, https://github.com/apache/skywalking-eyes/pkgs/container/skywalking-eyes%2Flicense-eye

